PR #27, opened 2014-12-19
PR/26

AndrewSmart:
Fixes 2 problems introduced in PR/25.

(1) command:pull can look at the **subrepo/$subdir** branch, which had been removed as it wasn't needed. It should have been looking at **subrepo/remote/$subdir** which had been fetched earlier by `fetch-subrepo`.

(2) In command:push, when GIT_SUBREPO_VERBOSE=true the `rebase -v` commands print:
```
fatal: ambiguous argument '': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

However the command succeeds anyway. Here is an example of the full output:
```
>>> git rebase -v --onto 0dd3e4e0c5c0a2890f9717a2c42e4e11fefb37db 1f312a6bd7a27f27959f04da8d6296a69452d34d subrepo/lib
fatal: ambiguous argument '': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Changes from  to 0dd3e4e0c5c0a2890f9717a2c42e4e11fefb37db:
First, rewinding head to replay your work on top of it...
Applying: p2 initial add to subrepo
```

I believe that has something to do with the porcelain command trying things during its fallback behaviour: http://stackoverflow.com/a/20967194/1185900

This patch removes that verbosity -v argument from the git rebase commands so that the misleading 'fatal' message is not printed.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -244,7 +244,7 @@ command:push() {
   subrepo-checkout
 
   # Attempt to rebase changes onto upstream:
-  FAIL=false RUN git rebase -v "$upstream_head" "$checkout_branch"
+  FAIL=false RUN git rebase "$upstream_head" "$checkout_branch"
 
   if ! OK; then
     # Let's rebase a bit more intelligently by using the log to see when the last clone/merge was.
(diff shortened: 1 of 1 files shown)


ingydotnet:
Applied. Thanks.

Sorry about the delay. Getting back to playing with subrepo and issues now.

AndrewSmart:
No problem. I think the issue this fixed wasn't shown in the test(s) due to the ordering of push/pulls (the push called subrepo:checkout so the later pull which had depeneded on the branch being checked out worked). I'll update ./test/failing/test.sh to show this in a different pull request (it's a trivial change though).

PR #29, opened 2015-01-07
PR/29 Example of PR/27 Fix

AndrewSmart:
Updates ./test/failing/test.sh to show what PR/27 fixed.
The issue is documented in that pull request (https://github.com/ingydotnet/git-subrepo/pull/27); this updated test exploits the issue that was fixed.
It's not necessary, but I thought the proof might be useful.

### DIFF
--- a/test/failing/test.sh
+++ b/test/failing/test.sh
@@ -50,6 +50,7 @@ rm -fr p1 p2 lib
   git add lib/readme
   git commit -m "p1 initial add to subrepo"
   git subrepo push --all
+  git branch -D subrepo/lib
 )
 
 (


ingydotnet:
Applied. Pushed.

I'm thinking on a refactoring of the pull process. I'll file an issue for that
later today.

ingydotnet:
@AndrewSmart I refactored this into a real test called test/issue29.t on the
`remix` branch. I got it to pass and it uses the new flow, but it looks a bit
suboptimal. I'd love to get your input on making it better.

fwiw, the new flow never involves a common ancestor, so that is why I think we
don't need those functions, but I could be wrong.

The new flow is to have a subrepo/<subdir> branch made with the `git subrepo
branch` command, and then rebase those commits onto an upstream ref made with
the `git subrepo fetch` command. This way the result is always pushable.

I have a bunch of out of sync subrepos over the last year that I will bring
into sync tonight and see how the new flow feels.

One thing is that when the `pull` fails now it undoes the entire rebase. It
might be better to keep the rebase intact, since it needs to be resolved
anyway. I think it would make this test case flow better, but not perfectly.

PR #63, opened 2015-02-01
Allows subrepo clone to clone to an empty branch; fixes #26.

### LINKED ISSUE #26: Cannot clone repo without making at least one commit
```
$ mkdir x
$ cd x
$ git init
$ git subrepo clone <url> -v
>>> git rev-parse --short HEAD
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Error: command failed: git rev-parse --abbrev-ref HEAD
```

If I touch a file and commit it, the clone works just fine. There shouldn't be a requirement of a first commit to use this plugin

AndrewSmart:
Changes:
* lib/git-subrepo:
  * subrepo:commit:
    * Must reset differently if $original_head_commit doesn't exist.
    * Index considers $subdir context as deleted, so I delete the index (not sure if there is a better way).
    * Respects the $GIT_INDEX_FILE environmental variable if defined.
    * **Creates a commit object without a parent as otherwise subrepo's history would be made the containing repo's history.**
  * assert-repo-is-ready:
    * Must not test based off of last commit if no last commit exists.
    * Tests detecting if working tree or index has changes must be different:
    * Working tree couldn't possibly have changes if there was nothing there to begin with.
    * **`git ls-files` non-empty output indicates index has changes.**
  * git:get-head-branch-name:
    * Now gets branch name when branch is empty.
* **test/subrepo-clone.t**:
  * Added testing for the subrepo clone into empty repo feature.

**Known Bugs:**
* Bug occurs on later pull after subrepo clone into empty repo. Bug is due to triggered subrepo.former check, as subrepo.parent is empty.
  * subrepo.parent is empty as the subrepo clone had no parent commit.
  * As soon as that deprecated subrepo.former check is removed this bug should be fixed.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -552,7 +552,17 @@ subrepo:commit() {
   GIT_WORK_TREE="$subdir" RUN git reset --hard "$subrepo_commit_ref"
 
   o "Reset index to original commit"
-  RUN git reset --mixed "$original_head_commit"
+  if [[ -n "$original_head_commit" ]]; then
+    RUN git reset --mixed "$original_head_commit"
+  else
+    RUN git reset --mixed
+    # index currently considers $subdir content as deleted at root.
+    # Delete the index
+    local indexfile="$GIT_INDEX_FILE"
+    [[ -n "$indexfile" ]] ||
+      indexfile="$(git rev-parse --git-dir)/index"
+    RUN rm "$indexfile"
+  fi
 
   o "Put info into '$subdir/.gitrepo' file."
   update-gitrepo-file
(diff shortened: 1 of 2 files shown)


AndrewSmart:
I changed a few things from that last pull request:
* I'd merged test/issue26.t into test/subrepo-clone.t
* Fixed the history bug.
* Changed the `git diff --cached` to `git ls-files` as the latter is faster.

Hmmm... I'll mark the changes bold above.

ingydotnet:
This looks good. I'm going to change it so that parent is set to "none" instead
of an empty string.

I think this will be more clear and less error prone.

ingydotnet:
Applied. Pushed. Thanks.

Did a bit of refactoring, but looks and works good. Set parent to 'none'. I
think it makes the .gitrepo look like it was done on purpose. Also should fix
the 'former' bug you mentioned.

PR #65, opened 2015-02-03
Prevents subrepo:commit from deleting untracked files in subrepo.

AndrewSmart:
Changes:
* lib/git-subrepo:
  * subrepo:commit:
    * Removes only tracked files in $subdir.
    * Ensures untracked files are not deleted or added to index.
    * Algorithm is to ignore untracked files:
      * Determine what is untracked (via git ls-files -o),
      * Add those to an ignore file via `$GIT_DIR/info/excludes` (alternatively could be via `core.excludesfile`),
      * Add everything as before with git add -A -- "$subdir",
      * Revert the ignore file.
    * Alternative algorithm would be to:
      * Record tracked files in two places (before/after `git reset --mixed`), only the union of the two works.
      * Must explicitly add .gitrepo as during initial clone it isn't tracked.
* test/subrepo-push.t
  * Adds a couple untracked files, and verifies that they exist after a pull/push.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -545,8 +545,15 @@ subrepo:commit() {
   o "Remove old content of the subdir."
   (
     cd "./$subdir"
-    RUN rm -fr $(ls -A)
+    RUN rm -f $(git ls-files)
   )
+
+  # "Ignore untracked files during work."
+  OUT=true RUN git rev-parse --git-dir
+  local git_dir="$output"
+  RUN cp "$git_dir/info/exclude" "$git_dir/info/exclude.bak"
+  OUT=true RUN git ls-files -o
+  echo "$output" >> "$git_dir/info/exclude"
 
   o "Put remote subrepo content into '$subdir/'."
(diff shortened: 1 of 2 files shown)


AndrewSmart:
Test fails on current master, but works with patch.

I think it would be a bit clearer to keep track of what files were untracked, and add those to $GIT_DIR/info/exclude during the `RUN git add -A "$subdir"` command.

As submitted it will break with spaces in the filenames; I think by reordering the string substitution commands it would work... I'll fix/redo this later, no more time tonight.

AndrewSmart:
Okay I fixed the bug I previously mentioned, where it wouldn't work with spaces in filenames.

AndrewSmart:
Oh crap, nevermind... still need to fix that.

AndrewSmart:
Ok, now it works with spaces in the file names. Had to add the untracked filenames to an ignore file temporarily.

ingydotnet:
Hey there @AndrewSmart . I'll be online tonight from 7pm-9pm PST. Join IRC if you can. I want to go over some core stuff.

grimmySwe:
Obsolete pull request.

Issue #69, opened 2015-03-06
Various problems with push


jrosdahl:
From documentation I get the impression that it should work to run `git subrepo push <subdir>` without specifying a subrepo branch, but I can't get it to work.

To reproduce what I'm trying to do, change

    git subrepo push bar subrepo/bar

to

    git subrepo push bar

in `test/subrepo-push.t`. The test case then fails like this:

```
test/subrepo-push.t .. 1/? #   Failed test 'push message is correct'
#   at test/subrepo-push.t line 45.
#     got: 'Subrepo 'bar' has no new commits to push.'
#   expected: 'Subrepo 'bar' pushed to '../../../tmp/upstream/bar' (master).'
test/subrepo-push.t .. 2/? #   Failed test 'File 'tmp/owner/bar/FooBar' exists'
#   at test/setup line 81.
test/subrepo-push.t .. Failed 2/6 subtests 
```

I guess it's supposed to work? If not, perhaps documentation could be improved?

Don't know if it's related, but another bug, I think, is that it doesn't work to push immediately after clone. This can be reproduced by removing the

```
(
  cd $OWNER/bar
  add-new-files bargy
  git push
)
```

part of `test/subrepo-push.t`. The test case then fails with

```
test/subrepo-push.t .. git-subrepo: 'subrepo/bar' is not a valid commit.
test/subrepo-push.t .. No subtests run 
```

instead of the expected

```
test/subrepo-push.t .. 1/? #   Failed test 'File 'tmp/owner/bar/bargy' exists'
```

I tried to make my own little standalone testcase, but it unfortunately only made me more confused. Here it is:

```
git init subproject
(
    cd subproject

    echo foo >foo
    git add foo
    git commit -m foo
)

git init superproject
(
    cd superproject

    # echo bar >bar
    # git add bar
    # git commit -m bar

    git subrepo clone ../subproject
    echo foo >>subproject/foo
    git commit -am foo

    git subrepo push subproject
)
```

The last command fails like this:

```
fatal: ambiguous argument 'none..HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Subrepo 'subproject' has no new commits to push.
```

When uncommenting the three commented lines, the failure is

```
The 'git rebase' command failed during your push.
You will need to finish the push by hand. Follow these steps:

  # Resolve the rebase conflicts
  git checkout master
[...]

PR #70, opened 2015-03-10
Fixes part of Issue #69; regarding where push following clone with 'none' $subrepo_parent (i.e. no initial commit before subrepo clone).

AndrewSmart:
Changes:
* lib/git-subrepo:
  * subrepo:branch
    * Adds in additional logic to handle the case where $subrepo_parent is 'none'.


grimmySwe:
Obsolete pull request, please test with latest master and see let us know if there is a problem.

PR #85, opened 2015-08-08
Add bash/zsh completions (subdirs, help, options)

perlpunk:
Add support for completing subrepo paths, help and options


ingydotnet:
Applied. Thanks!

Sorry took so long...

PR #90, opened 2015-09-02
fix #89: don't fail during push if there's nothing to push

### LINKED ISSUE #89: Impossible to push if there are several commits that don't contain subrepo changes.
For a host repository A with a subrepo B, if after a subrepo pull there are several commits in A that do not introduce changes to subrepo code, subrepo push will fail with a weird exception. Test and fix will follow.

dzzh:
If there are new commits in host repo, but nothing to push to subrepo, report that there's nothing to push instead of throwing an error.


grimmySwe:
Obsolete pull request, please test with latest master and see if this is still a problem.

PR #92, opened 2015-09-21
Fix #91 incorrect branching with merge commits

### LINKED ISSUE #91: Incorrect push when subrepo is filtered with a merge
When I do ````git subrepo push````, it does ````git subrepo branch```` as one of the first steps. The following steps are done when branching

- filter-branch with parent filter to discard commits prior to previous pull
- filter-branch with subdirectory filter to include subrepo commits
- filter-branch with parent filter to remove previous pull commit 
- filter-branch with tree filter to remove .gitrepo modifications 
- other changes

In one of my repositories I noticed the following when pushing. After the second step (subdirectory filter to include subrepo commits), subrepo commits were included as a separate branch that was merged to the filtered branch formed at the first step instead of forming a linear history. I'm not sure exactly why this happened, I was not able to reproduce it in the toy repository yet. After that, the third step has failed to remove the previous pull commit correctly and removed the oldest commit from the merged subrepo branch instead. This messed up the push pretty seriously and lead to like 15 old commits being committed to the subrepo again. 

The reason for this was the following. To find the commit to drop at step 3, you do ````git rev-list HEAD | tail -1```` .  This works fine if the branch has linear history, but if it contains merges, this returns an incorrect hash. To fix this, ````--first-parent```` option has to be added to ````git rev-list````.

You can see the situation at the screenshot. The green branch called subrepo-3 contains the changes since the last pull that I want to push (two topmost commits). Pink branch contains the changes from my subrepo that were added to the green branch with a merge commit at step 2. If we now execute ````git rev-list HEAD | tail -1````, it will return us not the oldest green commit (git subrepo pull...), but the oldest pink commit ([MAD-101] renamed...). 

screen shot 2015-09-21 at 14 42 13

dzzh:


grimmySwe:
Obsolete pull request, please test with latest trunk and let us know if it works.

Issue #96, opened 2015-10-12
Rebase is incorrect after branching


dzzh:
When working on #95, I found that when I do `git subrepo pull`, the newly created branch gets rebased incorrectly into `refs/subrepo/<subrepo_name>/fetch`. I have reproduced this in a test, will commit it soon.

In my test I have a subrepo that is called sub. After `git subrepo branch` is invoked, in my subrepo I have two files. The first is called `subrepo` with the first change from subrepo, the first one is called `subrepo-host` and contains changes introduced in my host repositort. 

Before rebasing my branch with these two files, in main subrepo repo I only have subrepo file with two changes from subrepo. 

When rebase command is invoked, I expect that it will complete successfully and will contain the following:
- `subrepo` file with two changes from subrepo repo
- `subrepo-host` file with changes from host repo

Instead I get a conflict. `subrepo-host` looks valid, for subrepo I get this.

```
direct change in sub
+<<<<<<< HEAD
+another direct change in sub
+=======
+>>>>>>> change from host
```

So my second change in `subrepo` file that is only present in subrepo repository conflicts with a change from host that was committed to a different file.

Git 2.4.6, OS X 10.10. 

Maybe it's better to not rebase after `git subrepo branch`, but rather squash all commits into one during  branching and then cherry-pick this single commit onto `refs/subrepo/<subrepo_name>/fetch`? If doing it this way, the previous pull commit should remain intact and contain all the previous history. So the newly created branch would look like

```
hgfedcba squashed changes after the last pull
abcdefgh git subrepo pull (and all preceding subrepo history)
```

dzzh:
I have implemented cherry-picking instead of rebasing and updated .gitrepo to contain separate references to pull parent and push parent. Now the application seem to work fine for my usage scenario. I'm not done with the fix, there are some issues with the code that can cause bugs and/or have to be otherwise improved. Do you think this can be integrated into the mainline after I complete the code and tests? If so, it will require to also write a migration script for the users installing the new version.

grimmySwe:
Test case for this issue is included in suite and it works, so I consider this issue closed.

Issue #95, opened 2015-10-09
Incorrect branch creation when there are merge commits for branches created before the previous pull


dzzh:
This issue seems to be related to #91. The fix I suggested there turned out to be a cure for a symptom, not disease.

`git subrepo branch` command doesn't work correctly when there's a merge of a branch created prior to the previous `git subrepo pull`. This can be reproduced as follows.

Imagine you have host and sub repositories (the latter is a subrepo of the former). Some developers commit directly to sub, some commit to host/sub. Every once in a while a host developer does `git subrepo pull` and `git subrepo push` to sync the changes. Now, consider the following scenario.

1. A host developer creates a feature branch from `master` branch in host to work on some new feature and commits to it.
2. Another developer does `git subrepo pull` in `master`.
3. Some people commit to `master`, their changes may contain subrepo updates.
4. Other people commit directly to sub.
5. The first host developer eventually completes his feature development and merges `feature` to `master`.
6. Another developer triggers `git subrepo pull` again... and we're screwed.

In this flow, when the last `git subrepo pull` is called, subrepo does its regular sequence of `git filter-branch` commands but as a result it gets a non-linear history. The feature branch that was created before the previous `git subrepo pull` execution and merged after the previous `git subrepo pull`, splits. The first it commits gets not only the changes made in this branch, but rather the whole snapshot of the working tree at time. This is not anticipated by the subrepo and it fails terribly when trying to do eventual rebase of `subrepo/sub` onto `refs/subrepo/sub/fetch`.

The best way to avoid this, I believe, is to rewrite the branching algorithm in subrepo. I'll try to come up with a solution.

I have a test script that reproduces the scenario I described earlier, I'll convert it to a test case and commit later today.

dzzh:
The referenced code fixed #96 and #95, but after applying it I cannot do push after pull, subrepo reports that there are no new commits to push. This can also be fixed, of course, but I'm not sure I'll find time for this soon.

AndrewSmart:
I ran your test for #95, you're right it fails. The last line of the test fails, on the rebase: [`RUN git rebase refs/subrepo/sub/fetch subrepo/sub`](https://github.com/ingydotnet/git-subrepo/blob/679ad12fd684bd72feff90d445c373d9a3d2aa0f/lib/git-subrepo#L442)
[...]

PR #97, opened 2015-10-15
issues 91, 95 and 96

dzzh:
This pull requests fixes a severe issue reported in #91, #95 and #96. Instead of rebasing during pulls and pushes I suggest to squash all the local changes into one commit and cherry-pick it onto upstream subrepo repository contents. This seems to be much more robust solution than rebasing.

Here I also separate pointers for pull parent commit and push parent commit. To do this, I had to change `.gitrepo` format. This means that after merging this pull request git-subrepo will become incompatible with earlier version. To overcome this issue, I can write a migration script. It will check whether .gitrepo has the old format and update it to the new one. Let me know if this is an acceptable solution.

### DIFF
--- a/Intro.pod
+++ b/Intro.pod
@@ -395,9 +395,9 @@ The pull command does these steps:
 
 =item * Create a branch of local subrepo commits since last pull
 
-=item * Rebase this branch onto the upstream commits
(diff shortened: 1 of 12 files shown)


grimmySwe:
This is an obsolete pull request, please check with latest master if your problems still occur and let us know.

PR #100, opened 2015-11-16
Issue98

grimmySwe:
Hi

Here is a suggestion on how to improve git-subrepo so it handles double pulls and don't lose changes so easily.

Instead of trying to find changes from last pull this one will take all changes in the subrepo and then try to find a common ancestor by using the referenced tree hash:es. It will then perform a rebase operation to move the two branches into same tree. Then it performs the regular rebase and in the push case it even uses -X theirs to avoid conflicts.

/grimmySwe


grimmySwe:
Included

Issue #106, opened 2015-12-03
Push to non-existent upstream branch messes up


ingydotnet:
Here is a replication strategy from memory:

* Clone ingydotnet/git-subrepo and checkout tag 0.3.0
* Make a fork of ingydotnet/test-more-bash
* Run this command:

    git subrepo push ext/test-more-bash \
        --remote=git@github.com/you/test-more-bash \
        --branch=new-test-branch

* You should a new branch on your fork with 40 commits
* Run subrepo push command again
  * You see 42 commits
* Run again
  * Command complains
* Run again
  * Command says up to date

Basically you can override the remote branch on a push, but if the branch
doesn't exist, it does it wrong.

ingydotnet:
Here's a quick look at what I see: https://gist.github.com/anonymous/b8e81fcbbfdc4348466f

grimmySwe:
@ingydotnet Let's see if get this correctly, summarizing the gist:
```
> git subrepo push -vd ext/test-more-bash/ --branch=issue106     # (1)
* Pushing to new upstream: git@github.com:ingydotnet/test-more-bash.git (issue106).
Subrepo 'ext/test-more-bash' pushed to 'git@github.com:ingydotnet/test-more-bash.git' (issue106).
> git subrepo push -vd ext/test-more-bash/ --branch=issue106     # (2)
Subrepo 'ext/test-more-bash' pushed to 'git@github.com:ingydotnet/test-more-bash.git' (issue106).
> git subrepo push -vd ext/test-more-bash/ --branch=issue106     # (3)
git-subrepo: Local branch is not updated, perform pull or use '--force' to always trust local branch in conflicts
> git subrepo push -vd ext/test-more-bash/ --branch=issue106     # (4)
git-subrepo: Local branch is not updated, perform pull or use '--force' to always trust local branch in conflicts
```
First (1) push will create the new branch. Second (2) should though not push new stuff, it should simply detect that there is no difference. 
(3) and (4) as as expected in this case as (2) has changed something upstream and you haven't pulled the changes back.

I'll see if I can create a simple test case for this.

grimmySwe:
This test case seems to show the problem:
```
source test/setup

use Test::More

clone-foo-and-bar

# Make various changes to the repos for testing subrepo push:
(
  # In the main repo:
  cd $OWNER/foo

  # Clone the subrepo into a subdir
  git subrepo clone ../../../$UPSTREAM/bar

  # Make a commit:
[...]

Issue #109, opened 2015-12-09
subrepo clone "doesn't contain upstream HEAD" on some tags


mmerickel:
First a clone using release tag 2.5.3:

```
git subrepo clone https://github.com/AFNetworking/AFNetworking.git foo -b 2.5.3 -v -d
* Fetch the upstream: https://github.com/AFNetworking/AFNetworking.git (2.5.3).
  * Fetch 'https://github.com/AFNetworking/AFNetworking.git' (2.5.3).
>>> git fetch --quiet https://github.com/AFNetworking/AFNetworking.git 2.5.3
  * Get the upstream subrepo HEAD commit.
>>> git rev-parse FETCH_HEAD
  * Create ref 'refs/subrepo/foo/fetch'.
* Make the directory 'foo/' for the clone.
>>> mkdir -p foo
* Commit the new 'foo/' content.
  * Check that '598eec5e1a1248f1bb313e5b7fcf9b6dfed00e01' exists.
  * Make sure '598eec5e1a1248f1bb313e5b7fcf9b6dfed00e01' contains the upstream HEAD.
  * Put remote subrepo content into 'foo/'.
>>> git read-tree --prefix=foo -u 598eec5e1a1248f1bb313e5b7fcf9b6dfed00e01
  * Put info into 'foo/.gitrepo' file.
>>> git config --file=foo/.gitrepo subrepo.remote https://github.com/AFNetworking/AFNetworking.git
>>> git config --file=foo/.gitrepo subrepo.branch 2.5.3
>>> git config --file=foo/.gitrepo subrepo.commit 598eec5e1a1248f1bb313e5b7fcf9b6dfed00e01
>>> git config --file=foo/.gitrepo subrepo.parent 0a22598cad6b6507544747fc99e1291f61503892
>>> git config --file=foo/.gitrepo subrepo.cmdver 0.3.0
>>> git add foo/.gitrepo
  * Commit to the 'foo' branch.
>>> git commit -m git subrepo clone --branch=2.5.3 https://github.com/AFNetworking/AFNetworking.git foo subrepo: subdir: "foo" merged: "598eec5" upstream: origin: "https://github.com/AFNetworking/AFNetworking.git" branch: "2.5.3" commit: "598eec5" git-subrepo: version: "0.3.0" origin: "https://github.com/ingydotnet/git-subrepo.git" commit: "c0fc373"
  * Create ref 'refs/subrepo/foo/commit'.
[...]

Issue #108, opened 2015-12-03
Reverted changes comes back


grimmySwe:
With the latest changes in 0.3.0 the current implementation is based on tree hashes for finding ancestors.

@jrosdahl Encountered a problem with this and there is a case where this fails:
Consider that you have a subrepo with tree hashes: `A->B->A`, in a main repo you have the tree hashes `A->B`. When you run pull you will actually find that the first subrepo A matches a previous A in the main repo, BUT you don't understand that it's actually connected to the previous A and the correct matching should be B to B. This pull will do nothing, treating the B as a valid change, when it should notice that B has been reverted into A.

grimmySwe:
After a lot of thought I think that the only viable solution is to define possible common ancestors as totally unique tree hashes in both branches. So If you find a common tree hash it can't be found anywhere else, because then we can't determine if a revert of some kind has been in progress.

Solution is to fail in these cases and let the user sort out what he/she wants.

I'll continue tomorrow with implementation of a solution with this I will also consider #107 as it's in the same ball park.

grimmySwe:
Implemented a solution for this on branch issue/107.

@ingydotnet During investigation of this issue I have consulted a lot with @jrosdahl (we do work at the same place :-) and there is some things to consider with this new tree hash solution.

It will be best where you interact or simply pull in. That way you will get a lot of common tree hashes.
It will be less good if you make a local change and never push it, that way you will never have common tree hashes and all operations are forced to use an older and older common ancestor. Worst case is that you only have the clone one in common.

The suggested workaround if you intend to have and keep local changes is to create a separate subrepo branch for that. And then perform the regular merges to that branch.

I have tried to come up with a solution to this but I see no easy way to do this.

ingydotnet:
Fixed by #110

Issue #107, opened 2015-12-03
Allow several pushes without pulls


grimmySwe:
Current workflow is that when you perform a push to a subrepo you must perform a pull to update your local `.gitrepo` file (if you don't supply the --update flag).
```
git subrepo push bar
git subrepo pull bar       <== You must do this otherwise it will complain that you haven't updated.
modify bar
git subrepo push bar
```
Consider the possibility of always using the ancestor check to see if you are allowed to push. That way it will be more similar to the regular git workflow where you don't have to perform pull after each push.
```
git subrepo push bar
modify bar
git subrepo push bar
```

grimmySwe:
Implemented a proposed solution on branch issue/107 (this also includes fixes for #106 and #108)

ingydotnet:
Fixed by #110

PR #110, opened 2015-12-09
Issue/107 + with fixes for #106, #107, #108, #109

grimmySwe:
Here some fixes for problems:
#106 #107 #108 #109

@ingydotnet Even if you don't do a new release it might be got to merge it to master so it easier to verify them and people that clone out master get the top branch.


ingydotnet:
@grimmySwe sorry to just be getting to this.

I've pushed this as issue/110 branch.

The one thing I'm not fond of is the --list option for the merge-base command. I
don't like is because it's  only for a single command.

Two possible solutions come to mind.

1) Use the `--all` flag instead since the normal meaning doesn't apply to
merge-base.

2) Use a special environment variable.

I'd go with the first one. In either case, you need to update the merge-base
doc in `doc/git-subrepo.swim`.

ingydotnet:
@grimmySwe, see 639a002 for a possible patch.

Also, I've lost track. What other issues would #110 close?

grimmySwe:
@ingydotnet I totally agree with your suggestion of using --all instead of --list, it is much better and consistent. 

Issues fixed in this is #106, #107, #108, #109, #110.

I am away on holiday so I can't perform any real tests but your suggested patch looks good to me.

ingydotnet:
Merged to master with fixes.

@grimmySwe, I'd like to make this into a release in early January. Give things
a review when you are back in action.

I'll close #106 #107 #108 #109 for now as well.

Issue #28, opened 2015-01-07
Add per command help like in git-hub


ingydotnet:
Help can be displayed on command usage errors too.

ingydotnet:
Fixed in #111

PR #111, opened 2015-12-13
add per command help

perlpunk:
see issue #28


ingydotnet:
@perlpunk, thanks for the PR. 2 things:

- It breaks test/subrepo-branch-all.t
- In `git-hub` when you type in the wrong cli args if shows the command help
  after the error.

I'll take a quick look and see if I can resolve them.

perlpunk:
oops, sorry, got the string comparison wrong.
`s/-ne/!=/`
pushed.

perlpunk:
i added the usage output in case of an error

ingydotnet:
Cool. I'll release with that.

I got the error messaging working but I think it is too noisy. Will create a
new issue, and push to a branch.

ingydotnet:
Applied. Pushed. Thanks!

I'll hold off on the error message stuff and start a new issue.

PR #114, opened 2015-12-24
show usage only for usage errors

perlpunk:
added a function usage-error()


ingydotnet:
Applied. Pushed. Thanks!

Issue #122, opened 2016-01-22
subrepo init broken?


nevsan:
Maybe I don't understand how I'm supposed to use `git subrepo init`, but I tried the following:

1. Run `git subrepo init <subdir> -r <myremote> -b master` from within an existing repository. This creates a new commit that simply adds the `<subdir>/.gitrepo` file.
2. Run `git bare --init` on a repo at the location `<myremote>`.
3. Run `git subrepo push <subdir>`.
4. Wonder why I got the error `git-subrepo: Can't find a common ancestor between refs/subrepo/<subdir>/fetch and subrepo-push/<subdir>. This might be the case if someone used push --force on the subrepo. You need to find out why and possibly reclone the subrepo.`

I notice that the remote repository does in fact receive the push (it appears in the history of the bare repository). It just seems that I can't push or pull from the repository once it has been created this way. I've confirmed that another fresh project can subrepo clone it and successfully push and pull to the remote--it's only the repo from which I initially ran subrepo init that I cannot push or pull.

I've traced the problem down to `subrepo:merge-base` and `subrepo:branch`. When doing a subrepo push or pull, it looks for a /unique/ merge-base. Note that `subrepo init` created a commit that only added the `.gitrepo` file. However, `subrepo:branch` will delete this file when generating the shadow history. Therefore, you will end up with two commits that have the same tree hash (the commit created by `git subrepo init` and that commit's parent). `subrepo:merge-base` notices that there are multiple candidates and complains, causing `subrepo:push/pull` to fail.

I've only gotten started trying to use this tool, so I don't know what the right solution is here. Perhaps `subrepo:merge-base` should check for this case explicitly, ignoring any commits that have the `git-subrepo init` metadata.

grimmySwe:
Hi Nevada

In step 2 I assume that you intend to `git init --bare`, because I don't have the `git bare` command ;-)

And yes, you seem to have found an issue here and pinpointed the problem. I tested myself and as you say it looks like when you use `git subrepo init` you will get a duplicate commit from a tree hash perspective and that will make merge-base go crazy.

I looked at you suggested patch and it will solve this specific case, but I stumbled on another case where I accidentally updated only the .gitrepo in another commit and this will cause the same problem. 

When I look at the result created by `git subrepo branch` I see that commits only changing the .gitrepo file is still present and that is strange.

I need to investigate some more, but main main thesis right now would be to clean the log so that only commits actually changing stuff in the subrepo should be present.

[...]

PR #123, opened 2016-01-22
Skip commits created by 'git subrepo init' when finding merge-base.

nevsan:
Possible solution to #122 

This removes commits created by `git subrepo init` as candidates for `git subrepo merge-base`.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -739,7 +739,7 @@ subrepo:merge-base() {
     first_on_target="HEAD"
     for target in $second; do
       local array2=(${target//:/ })
-      if [[ $tree == ${array2[1]} ]]; then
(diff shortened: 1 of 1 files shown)


grimmySwe:
I think that this solution will solve one specific case, but that there is a bigger problem left untouched.

ingydotnet:
This should probably have a test.

grimmySwe:
I investigated this and found a solution that I think solved this in a more
complete way by using prune-empty to parent-filter. Suggestion is in
issue/122, currently on review to Nevada the reporter.

fredag 22 januari 2016 skrev Ingy döt Net <notifications@github.com>:

grimmySwe:
Closing this, solved by #127

Issue #124, opened 2016-01-22
Can't run git subrepo commands from the top level of a submodule


nevsan:
`git subrepo` commands fail at the top level of a new-style submodule because it thinks that it is not at the top level of a git repository. I believe that a submodule should be treated as a standalone repository from the perspective of git-subrepo.

The problem comes from the use of `git rev-parse --git-dir == .git` as a method for determining whether one is at the top level of a repository. For new style submodules, the .git directory lives in the parent repository's .git directory.

grimmySwe:
To me your suggested solution looks good, could you please supply some test cases as well?

PR #125, opened 2016-01-22
Use 'git rev-parse --show-prefix' to test if at top level.

nevsan:
This check lets git-subrepo work within submodules (solving #124 )


ingydotnet:
Applied. Thanks!

This gets me thinking about how to make subrepo work well from any subdir.

The tricky part is that the subrepo "names" are the same as the relative path
when you are at the top level.

ie If you have a subrepo called `foo/bar/baz` and you are cd'd to `foo/bar`,
you'll want `git subrepo status baz` to work.

PR #127, opened 2016-01-22
Add prune-empty to the filter-branch command.

grimmySwe:
This should solve #122 and make the branch command much better. It will clear out all empty commits caused by the `filter-branch --tree-filter`command. They could cause trouble for other cases then init.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -678,7 +678,7 @@ subrepo:branch() {
 
   o "Remove the .gitrepo file from the history."
   FAIL=false RUN git filter-branch -f \
-    --tree-filter "rm -f .gitrepo"
+    --prune-empty --tree-filter "rm -f .gitrepo"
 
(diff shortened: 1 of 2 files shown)


grimmySwe:
@ingydotnet  Hmm, that didn't go as I expected, I actually want this pull request to trunk... Commit: f172780
If I create a pull request to master from issue/122 I get some extra stuff already included, so I created it from the base branch (issue/115)...

ingydotnet:
@grimmySwe, I fetched your PR and rebased it and pushed the result to
https://github.com/ingydotnet/git-subrepo/tree/issue/127

Let me know if I can use that, or if not, what should I do next.

grimmySwe:
@ingydotnet Sorry for the extra work, still not used to git/github... I checked and it looks like you could merge in issue/127 to master.

ingydotnet:
Applied. Thanks!

PR #138, opened 2016-02-08
Issue/130 - Faster algorithm for common tree ancestor

grimmySwe:
Implemented a faster algorithm to find common tree ancestor.

(S=Subrepo, C=Clone of subrepo)
Previous best case: S+C, worst case S*C
new algorithm has best case S+C, worst case of 2S+C

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -84,6 +84,8 @@ main() {
   local new_parent=             # New subrepo parent commit
   local first_on_source="HEAD"  # Merge base source is HEAD
   local first_on_target="HEAD"  # Merge base target is HEAD
+  local merge_counter=0         # Unique prefix for merge bases
+  local branch_first_commit=    # Return value from prepare_merge_base
 
   local subdir=                 # Subdirectory of the subrepo being used
   local gitrepo=                # Path to .gitrepo file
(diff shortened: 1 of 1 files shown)


ingydotnet:
I have no problem merging this. Just wondering...

* Are the existing tests good enough to catch regression?
* Could this be faster in Perl or Python?
  * Both are commonly installed on modern OSes
  * We could shell out if available, else use Bash

Regarding caching the last result, I'm a bit leery. Where would we store it?
Might it possibly go stale? It would certainly need to be per branch, since
every branch of a repo can very in subrepo usage/state.

But back to merging this. I assume this is a step forward, with no known
worries of regression, right?

grimmySwe:
It actually caught a problem with pull --all. So I consider it "safe"...

Regarding speed I don't really know, but I think that the complexity for having external dependencies should be avoided until we really see the problem. And it might be a little deal breaker for windows.

So to conclude, yes this is a step forward and I think it should be ok to merge.

For the suggestion on caching I have come to the same conclusion as you, it might not be worth the risk. In git you can do all kinds of things like rebase and then push --force leaving caches invalid and hard to detect.

ingydotnet:
Applied. Pushed. Thanks!

Issue #148, opened 2016-02-15
Completion for `init` should complete to subdirs


ingydotnet:
not to subrepos

Attn @perlpunk

ingydotnet:
This seems resolved. Closing.

PR #149, opened 2016-02-15
subrepo init should complete to directories, not existing subrepos

perlpunk:
see #148


ingydotnet:
@perlpunk, this works but it completes to files as well as directories. It
should really only complete to non-empty directories. And maybe also only to
dirs without a .gitrepo subdir.

perlpunk:
This might take a while. Right now it's just using the default completion for bash, which is file completion.
I'd have to reimplement the builtin file completion.
You might want to merge this for now and create a new issue...

ingydotnet:
sounds good

ingydotnet:
Applied. Pushed. Thanks!

PR #151, opened 2016-02-19
Proper GIT_SUBREPO_COMMAND format under Environment Variables

maxmclau:
It wasn't recognizing it as code because of a missing `back-tick and it was bothering me


ingydotnet:
Thanks for noticing, but...

The ReadMe is generated. The doc sources are in `doc/`.

    make update

Generates the generated docs.

Anyway... fixed now. Thanks again!

Issue #136, opened 2016-01-28
Make status work different inside subrepo branch


ingydotnet:
When a subrepo branch is checked out, the status command show useful info, like
rebase status and previous branch.

It might be able to find the .gitrepo file info if it can accurately identify
the previous branch; possibly from reflog.

See #37

Issue #132, opened 2016-01-27
Support `--all` for status command.


ingydotnet:
This will show all the nested subrepos in addition to the top level ones.

See #82

stsquared99:
Should this really be an '--all' flag? It would be inconsistent with the function of push --all or fetch --all, for example. Unless we also modify those commands to work for all nested subrepos as well, but I'm not sure we want that. Perhaps a --recursive flag?

ingydotnet:
@stsquared99 I realize this, but: https://github.com/ingydotnet/git-subrepo/issues/82#issuecomment-175363231

For better or worse, I've set things up so that all the commands share the same options. We don't have any options that only apply to a single command. I'm ok with 'all' meaning something different for status, show and clean.

ingydotnet:
@stsquared99 I had a think on this, and you are probably right. We should have consistency for `--all`.

I'd like to propose a `--ALL` (`-A`) flag for subsubrepos. This can be applied to the `status`, `show` and `clean` commands at a minimum.

I'll go ahead and add it for `clean` and then you can use it for `status` and `show`?

FWIW, `--all` should still be the default for `status` and `show`.

stsquared99:
Hmm, I see. I think that works fairly nicely. I'll resend my latest pull when you have that added in.
Should we try to add -A to the rest of the commands as well?

ingydotnet:
I'd rather not, until there is a real need for it. People can always do the other commands explicitly per subsubrepo.

ingydotnet:
Pushed `--ALL` support to branch `issue/132b`. Please test out. Need to doc it, and make a test for it.

stsquared99:
Thanks, that logic looks much better than what I originally was going to do

ingydotnet:
Done. Closing.

Issue #84, opened 2015-07-24
`status` command shows too much with `-vF`


ingydotnet:
The `-v` option for `status` shows ref information. The `-v` for other commands
shows internal steps. This is all correct.

The `-F` flag for `status` fetches ref info, but when used with `-v` errantly
shows the steps too.

stsquared99:
Should this be done, considering our decision in regards to --all vs --ALL? The code for --verbose is very clean right now, we would have to add an exception. The ref information isn't very much, perhaps that should just be printed by default?

ingydotnet:
I'm not sure. I played around with it and I like the -v status vs normal and -q.

Let's wait on this for now.

stsquared99:
I agree, I do like that functionality as well, it just seems awkward in function and in code to change the -v behavior.

In the meantime, can I send you a PR for the show vs status #163, or should I wait until the rest of the current status issues are resolved?

ingydotnet:
Please go ahead with #163. Thanks!

Issue #82, opened 2015-07-16
nested subrepos


ecraven:
Hello :)

I'm working on a php project, I have a core, and inside the core, in various nested directories, plugins. I'd like to have one git branch for the core, and one for each plugin.
If I try to clone them with git subrepo into nested subdirectories, strange things happen:
````
git subrepo clone /path/to/core core
git subrepo clone /path/to/plugin core/plugin
git subrepo status
````
This now shows *one* subrepo, core (although the actual data is checked out and present).
How do I properly nest subrepos? I don't want to include plugin in the core repo, but I want one "working" repo that has all the plugins inside core in various places. 

Thanks for any help,

Peter

ingydotnet:
Hi @ecraven. Thanks for the report.

Subrepos were made so that you could `git subrepo clone` repos, that themselves
had subrepos. It was not intended that you could clone subrepos under a subdir
that was already a subrepo. I should patch to not allow that.

If you run this command on the `ingydotnet/git-subrepo` itself:

    find . -name .gitrepo

you'll see that there are nested subrepos! There are 4 `.gitrepo` files in all.
But then run:

    git subrepo status

And you'll only see 2. That's because `git-subrepo` repo only cloned in 2
subrepos, but one of them had 2 subrepos itself.

This is the intended behavior.

It sounds like you are working on WordPress? WordPress is an area I would like
to see make use of subrepos, and probably for the reasons you want. So I'd like
to help you get happy.

One simple solution might be this:

    git subrepo clone /path/to/plugin plugins/plugin
    ln -s core/plugin ../plugins/plugin

In other words, put the plugin subrepos in their own `plugins` subdir and then
figure out a way so that the core can find them. Maybe using symlinks, maybe
something else.

Let me know what you think.

ingydotnet:
I've been thinking on this and my mental model has shifted…

I used to think that sub-subrepos should be invisible to the parent repo, but
now I think we should expose them as far as the credentials of the author allow
them to go.

In other words, if I am the owner or collaborator on a subrepo of some subrepo,
I should be push directly to its upstream.

I should also see the sub-subrepos (and sub-sub-subrepos etc) in a status
command. Although I suspect I'll want to make that aspect clear, perhaps by
indentation.

I'll start the `subsubrepo` branch for this work.

grimmySwe:
@ingydotnet 
[...]

PR #161, opened 2016-03-10
Fixes a bug that occurs when attempting to created nested subdirs

stsquared99:
Fixes a bug that occurs when trying to create a subrepo inside of another subrepo.

###

Steps to reproduce:

$ find | grep -v '\.git'
.
./foo
./foo/bar
./foo/bar/Bar
./foo/Foo

$ git subrepo init foo
Subrepo created from 'foo' (with no remote).

$ git branch
* master
  subrepo/foo

$ git subrepo init foo/bar
git-subrepo: Command failed: 'git branch subrepo/foo/bar'.

$ find | grep -v '\.git'
.
./Bar

###

'git branch subrepo/foo/bar' fails because '.git/refs/heads/subrepo/foo' already exists as a file instead of a directory. Furthermore, when git branch fails, there is no catch, so the HEAD is stuck in the subdirectory-filter.

This PR properly handles the failed 'git branch' command, and prevents a subdir from being created inside of another subdir entirely.
Hopefully the test I wrote is in accordance with your standards.

Thanks

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -485,7 +485,11 @@ subrepo:init() {
 
   o "Create branch '$branch_name' for this new subrepo."
   FAIL=false RUN git branch -D "$branch_name"
-  RUN git branch "$branch_name"
+  FAIL=false RUN git branch "$branch_name"
+  if ! $OK; then
+    RUN git reset --hard "$original_head_commit"
+    error "Failed to 'git branch $branch_name'."
+  fi
 
   if [[ $subrepo_remote != none ]]; then
     o "Create subrepo remote 'subrepo/$subdir' -> '$subrepo_remote'."
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -1344,7 +1344,7 @@ assert-subdir-ready-for-init() {
     parentdir="$parentdir$dir/"
     if [[ -e $parentdir.gitrepo ]]; then
       error "The subdir '$subdir' is already part of another subrepo at "\
-"'$parentdir'."
+"'${parentdir%/}'."
     fi
   done
   # Check that subdir is part of the repo
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -1338,6 +1338,15 @@ assert-subdir-ready-for-init() {
   if [[ -e $subdir/.gitrepo ]]; then
     error "The subdir '$subdir' is already a subrepo."
   fi
+  # Check if subdir is already part of another subrepo.
+  parentdir=
+  for dir in $(echo "$subdir" | tr '/' '\n'); do
+    parentdir="$parentdir$dir/"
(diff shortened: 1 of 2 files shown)


ingydotnet:
Great PR. Thanks. Only problem is I'm not sure there is a problem.

I can run:

```
git subrepo init ext/bashplus/test
```

in the git-subrepo directory itself and it works fine.

Should we prevent init-ing subrepos inside subrepos? I'm not sure.

stsquared99:
The reason it works is because the 'subrepo/ext/bashplus' branch doesn't exist yet in your local repo. If you run 'git subrepo branch ext/bashplus', pull or push with processed commits, or run 'git subrepo init ext/bashplus' for the first time, the branch will then exist, which will block 'git subrepo init ext/bashplus/test'. It took me a while to realize this myself while I was trying to trace the issue.

Try running 'git subrepo branch ext/bashplus', then run 'git subrepo init ext/bashplus/test' and you will see the problem.

Thanks

ingydotnet:
Thanks for the explanation. BTW, I like working with existing public repos to
recreate these issues and replication strategies.

The thing is, that with your patch I can't init ext/bashplus/test and I think
that I want that to be possible. I ran into a situation where I was writing
code in a subdir of some project and I started thinking, this will be a subrepo
later, and parts of it might become subrepos too.

There are still a few open issues around subsubrepos. I'd like to get those
resolved with tests to keep them from regressing.

I'll add a test to protect my use case above and then you can adjust your patch
to make it work under that constraint.

stsquared99:
I see, that makes sense. It is rather restrictive to not allow any subrepos you are using to not have subrepos of their own.
The only thing standing in the way of that is the branch names that are being used, due to the dir vs file problem (subrepo/foo & subrepo/foo/bar, foo being a file in the former case and a dir in the latter). If we can find a way around that, then subrepos can be nested without a problem.

Perhaps we should move this discussion to an issue?

In the meantime, I will send you a separate pull with just the commit that catches and resets after a failed git branch.
Thanks

ingydotnet:
@stsquared99 see https://gist.github.com/2b1c142254fdf83a114b

git-subrepo already has a sub-subrepo, as you can see.

The thing is that the status command can't see them yet (unless you specify them explicitly).

See #132. If you could work on that one too, I'd appreciate it. Else I will eventually...

ingydotnet:
#136 and #84 are the other open status related issues. Just FYI...

ingydotnet:
Also #82 is the nested subrepo issue.

Regarding the dir vs file thing, I think just issuing a message to do a `clean` command first when conflicts happen should be enough.

stsquared99:
Ah, I think that clarified things for me now, thanks. I'll take a look at those issues and play around with nested subrepos some more.

PR #164, opened 2016-03-14
Catch errors when creating a subrepo branch

stsquared99:
Related to #82, but will prevent being stuck in the subdirectory-filter no matter what causes 'git branch' to fail.


ingydotnet:
How does the branch command fail?

stsquared99:
Right now, I see it when doing a 'init' or 'branch' in a nested subrepo when a parent subrepo branch exists, or vice versa, due to the dir/file thing we discussed earlier. I am uncertain whether there are/aren't any other cases where it could fail.

stsquared99:
No longer an issue with #173

PR #190, opened 2016-05-26
Loosen up regex to determine if pushing to a new repo after init

loupgaroublond:
The push command expects the "fatal: Couldn't find remote ref" string to be present. However, all manner of various outputs from git could fail to match that string at the beginning of the output. This relaxes this restriction with the addde bonus that the commit message is longer than the change.

Example:
✔ ~/BitPusher/bookstore/bp-automation-merge [cluster/devops ↓·18↑·3|✔] 
17:55 $ git subrepo push chef/cookbooks/bp-aws-single-az-lb --force
git-subrepo: Fetch for push failed: Host key fingerprint is SHA256:SCBz9Sh537ZPrxSnRQRymcFY7W6FazDAhvTcwWoJiEM
+---[RSA 2048]----+
|  o.Eo.o.o.=**.  |
|   +oo.ooo=+=o.  |
|    o.+ .oo+.... |
|     + o .+ o.o .|
|      . S.o .+oo |
|         . . ==  |
|          . +o   |
|           + .   |
|            o..  |
+----[SHA256]-----+

User ynemoy authorized.
fatal: Couldn't find remote ref master
fatal: The remote end hung up unexpectedly

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -578,7 +578,7 @@ subrepo:push() {
     if ! OK; then
       # Check if we are pushing to a new upstream repo (or branch) and just
       # push the commit directly. This is common after a `git subrepo init`:
-      local re="^fatal: Couldn't find remote ref "
+      local re="fatal: Couldn't find remote ref "
       if [[ $output =~ $re ]]; then
         o "Pushing to new upstream: $subrepo_remote ($subrepo_branch)."
         RUN git push \


ingydotnet:
@loupgaroublond Thanks for this and sorry for late reply.

I tightened up the regex to make sure text is at start of line. I've pushed
this to branch `issue/190`. It passes the tests (and the regex check is covered
in the tests).

@loupgaroublond, would you mind verifying it?

loupgaroublond:
Hi Ingy,

I can't test this in the same environment anymore as I am not working on that project. As long as it can handle multiline output from SSH and git, I think the fix should work. I plan on using git-subrepo soon on a new project, so if I run into this issue again, ill let you know.

ingydotnet:
Applied. Pushed. Thanks.

I've pushed this fix. I should note there's no test for it. I'll make a note to
add a test but I want to spend time getting more issues resolved.

I tested the regex I had and realized it didn't work, so I changed it to one
that does. Tests pass, even though it doesn't tell us anything about this. :)

Thanks again, Ingy

PR #204, opened 2016-07-10
Fixed #203: added support in Makefile for DESTDIR variable.

### LINKED ISSUE #203: Support staged installs via the DESTDIR variable
The `DESTDIR` variable is a common pattern supported only in install\* and uninstall\* targets of a Makefile often used in package creation or other fakeroot installation situations. The specifics of the `DESTDIR` variable can be found in the [GNU Make Manual](https://www.gnu.org/software/make/manual/html_node/DESTDIR).

`DESTDIR` should never be defined in the Makefile, but, when specified as an absolute path by the user when invoking the make command, its value should be prepended to each install target file. For example, if the invocation, `make install`, would normally install the target files: `/usr/local/bin/foo` and `/usr/local/share/man/man1/foo.1`, executing `make DESTDIR=/tmp/stage install` should likewise install these corresponding target files: `/tmp/stage/usr/local/bin/foo` and `/tmp/stage/usr/local/share/man/man1/foo.1`.

jamesan:
I'm packaging git-subrepo for Arch Linux at the moment and this change works insofar that use case. The Arch Linux package can be found at: https://aur.archlinux.org/packages/git-subrepo-git.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,9 +15,9 @@ SHARE = share
 
 # Install variables:
 PREFIX ?= /usr/local
-INSTALL_LIB  ?= $(shell git --exec-path)
+INSTALL_LIB  ?= $(DESTDIR)$(shell git --exec-path)
 INSTALL_EXT  ?= $(INSTALL_LIB)/$(NAME).d
-INSTALL_MAN1 ?= $(PREFIX)/share/man/man1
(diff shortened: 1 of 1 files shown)


ingydotnet:
Applied. Pushed. Thanks!

Sorry for the delay.

THe latest greatest v0.4.0 should be coming out within a week or so. FYI.

Issue #206, opened 2016-07-18
Subrepo hooks documentation


jmatsushita:
Would be great to have examples of how to setup for instance a pre-push hook on several subrepos.

jmatsushita:
Doing a `git subrepo push` with the `pre-push`:
```
#!/bin/sh
echo $GIT_SUBREPO_RUNNING
echo $GIT_SUBREPO_COMMAND
exit 1
``` 

Nothing is printed even with `-d` or `-v`. It does prints empty strings when using `git push`.

jmatsushita:
Ha `-x` does display the $PID and `push`.

jmatsushita:
Hmm... Looks like it would be useful if the `<subdir>` was also exported. Or if GIT_SUBREPO_COMMAND also exported the subrepo arguments ?

jmatsushita:
It seems I can work around and find the `subdir` by using the `while read local_ref...` 

Next problem, I can't seem to be able to run an interactive command (I'm trying to setup a hook that will request interactively to upgrade the semantic version of specific projects).

I'm very far from being competent at bash, but when trying a few things (particularly starting a bash -i from the hook) I get `bash:no job control in this shell` which seems to indicate that the parent script might not allow interactive subscripts?

In the end I try to use something like:
```
exec < /dev/tty
(
interactive_command
)
exec <&-
```

Which works fine with `git push` but just outputs the result with `git subrepo push -x` without waiting for input.

devinrhode2:
@jmatsushita Any news on using semver with git alone, and not with npm? (and not necessarily with git-subrepo either)

PR #207, opened 2016-07-18
[doc] GIT_SUBREPO_COMMAND  is set to name not pid

jmatsushita:
#206


ingydotnet:
See: https://github.com/ingydotnet/git-subrepo/blame/master/ReadMe.pod#L4

The ReadMe is generated.

I updated your commit to patch the doc properly. Thanks for submitting it.

Issue #211, opened 2016-07-21
git-subrepo clone doesn't work with email as part of url for authentication


gcstang:
git-subrepo clone ssh://user.name%40domain.com@git.mydomain.com:3333/subreposub1.git subreposub1
Returns
git-subrepo: Command failed: 'git ls-remote ssh://user.name domain.com@git.mydomain.com:3333/subreposub1.git'.

perlpunk:
Right, the percent sign is interpreted by printf.
Can you try out if pull request #212 works for you?

gcstang:
That worked!

Thank you

grimmySwe:
This issue is solved and available on master.

PR #212, opened 2016-07-21
escape percent signs in arguments to printf

perlpunk:
see issue #211

```
git subrepo clone ssh://user.name%40domain.com@git.mydomain.com:3333/subreposub1.git
```


perlpunk:
rebased to current master because of conflict with the latest commit

ingydotnet:
Applied. Pushed. Thanks!

Issue #192, opened 2016-05-31
git subrepo init fails


kilowu:
Repo has a sub-directory named java.
$ git subrepo init java
git-subrepo/lib/git-subrepo: line 1174: printf: `+subdir': not a valid identifier

grimmySwe:
Could you please add some more information for this case:
What subrepo version are you using? 
Can you reproduce this error? How? Can you create a simple example to reproduce it?

kilowu:
I'm using the latest version of subrepo (pulled from master).

I found out that the problem is about to the bash environment. 

There is a [[ command's =~ operator in git-subrepo line 1166:
```bash
   if [[ $arg =~ ^\+ ]]; then
      if [[ $i -ge $num ]]; then
        usage-error "Command '$command' requires arg '${arg#+}'."
      fi
      printf -v ${arg#+} "${command_arguments[i]}"
    # Look for function name after ':' to provide a default value
    else
      if [[ $i -lt $num ]]; then
        printf -v ${arg%:*} "${command_arguments[i]}"
      elif [[ $arg =~ : ]]; then
        "${arg#*:}"
      fi
    fi
```

The rhs argument is ``^\+``. It does not match the "+pattern" in my machine. I tried to change it to ``^\\+`` and after that it worked in my environment.

It looks like a bash compatibility issue.

hainm:
> The rhs argument is ^\+. It does not match the "+pattern" in my machine. I tried to change it to ^\\+ and after that it worked in my environment.

I got the same issue with current master branch (1e795953d4265abacf828ed7e66036cff281fc7c) and your solution works for me too.

ingydotnet:
@kilowu @hainm could you both post the output of:

```
uname -a && echo && bash --version
```

hainm:
mine
```bash
Linux [...] Wed Sep 3 16:22:34 EDT 2014 x86_64 x86_64 x86_64 GNU/Linux

GNU bash, version 3.2.25(1)-release (x86_64-redhat-linux-gnu)
Copyright (C) 2005 Free Software Foundation, Inc.
```

ingydotnet:
@hainm which linux distro and distro version?

hainm:
```bash
$ cat /proc/version 
Linux version 2.6.18-371.12.1.el5 (mockbuild@builder17.centos.org) 
(gcc version 4.1.2 20080704 (Red Hat 
[...]

PR #214, opened 2016-08-23
Fix regex escape compatibility issue #192

kilowu:
Using string comparison operator instead of regex matching which avoids the compatibility problem in different shell environments.


ingydotnet:
Applied. Pushed. Thanks.

Much cleaner approach.

PR #219, opened 2016-08-30
Force .gitrepo add.

mdriesch:
If the container repo has a .gitignore which contains a rule that, possibly by chance, ignores the .gitrepo file, then the subrepo clone command fails. The `add -f` (force) fixes the issue.


ingydotnet:
Applied. Pushed. Thanks!

I added a comment to write a test for this someday.

Issue #215, opened 2016-08-25
Updating subrepo failure: git-subrepo: Command failed: git rebase --onto


micah:
i'm having trouble updating a subrepo, i did this:

```
$ git subrepo pull puppet/modules/tor
git-subrepo: Command failed: 'git rebase --onto dcb6e748864e7dfd3c14f4f2aba4c9120f12b78a 429cb2d55be683e540960ddd3ad175dc97ec03a0_b37ab7ad1bb2de098791f557355358fe24177f49_7a1872d32c698406bdb3566cfa768419dd447cdc subrepo/puppet/modules/tor'
```

if I run with debug to see the commands:

```
$ git subrepo -d pull puppet/modules/tor
>>> git rev-parse --verify HEAD
>>> git fetch --quiet https://leap.se/git/puppet_tor master
>>> git rev-parse FETCH_HEAD^0
>>> git branch -D subrepo/puppet/modules/tor
>>> git filter-branch -f --subdirectory-filter puppet/modules/tor HEAD
>>> git filter-branch -f --prune-empty --tree-filter rm -f .gitrepo
>>> git branch subrepo/puppet/modules/tor
>>> git reset --hard b928a8f29b40df545ac6a72ead6cb9ed6a36fbda
>>> git rebase --onto dcb6e748864e7dfd3c14f4f2aba4c9120f12b78a 429cb2d55be683e540960ddd3ad175dc97ec03a0_b37ab7ad1bb2de098791f557355358fe24177f49_7a1872d32c698406bdb3566cfa768419dd447cdc subrepo/puppet/modules/tor
git-subrepo: Command failed: 'git rebase --onto dcb6e748864e7dfd3c14f4f2aba4c9120f12b78a 429cb2d55be683e540960ddd3ad175dc97ec03a0_b37ab7ad1bb2de098791f557355358fe24177f49_7a1872d32c698406bdb3566cfa768419dd447cdc subrepo/puppet/modules/tor'.
$ 
```

If I run the commands one by one, when I get to the rebase one at the end, I'm told:

```
fatal: Needed a single revision
invalid upstream 429cb2d55be683e540960ddd3ad175dc97ec03a0_b37ab7ad1bb2de098791f557355358fe24177f49_7a1872d32c698406bdb3566cfa768419dd447cdc
```

Any ideas?

ingydotnet:
@grimmySwe what's with that invalid rebase sha with underscores in it?

[...]

PR #226, opened 2016-09-20
Fix length comparison operators in `merge-base`

twooster:
I'm not entirely clear on when this will be necessary (not enough time to trace the code), but it's the state I found my repo in. If anyone who understands the context better wants to supply a regression test, or even supply context so I can write a regression test, be my guest.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -809,7 +809,7 @@ subrepo:merge-base() {
   for entry in ${merge_base_order[@]}; do
     local source_key="${source}_$entry"
     local source_value="${!source_key}"
-    if [[ ${#source_value} > 40 ]]; then
+    if [[ ${#source_value} -gt 40 ]]; then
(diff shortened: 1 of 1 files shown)


lib/git-subrepo
```diff
@@ -809,7 +809,7 @@ subrepo:merge-base() {
   for entry in ${merge_base_order[@]}; do
     local source_key="${source}_$entry"
     local source_value="${!source_key}"
-    if [[ ${#source_value} > 40 ]]; then
+    if [[ ${#source_value} -gt 40 ]]; then
```
gzm55: did you try ' (( ${#source_value} > 40 ))' ?

twooster:
Did you?

Sorry, but what's the point of the question? Both work.

ingydotnet:
This was also fixed in #215 which I just applied. Should be all good. Thanks!

PR #229, opened 2016-09-29
Update .rc to work with installation paths with spaces

lcorbasson:


ingydotnet:
Applied. Pushed. Thanks.

Actually stuff was overquoted. Some by me and some by you. I took out the
unnecessary quotes and checked it with `shellcheck`. shellcheck is an excellent
too for things like this. At some point I'll apply it to the entire code base.

Note, if this doesn't actually work as intended, please reopen. Cheers.

PR #230, opened 2016-10-05
pulling a new branch doesn't update the branch name

2improveIT:
Fix pulling a new branch which the same commit doesn't update the branchname in the .gitrepo file


ingydotnet:
Great fix. Applied. Pushed. Thanks!

I refactored the code a bit to conform more to `shellcheck`.

Thanks for taking the time to add a test!

PR #234, opened 2016-10-24
Don't reuse previous commit message when using --all

jrosdahl:
Before this fix, all commits for pulled subrepos would get the same message (created for the first repo) when using the -a/--all option.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -120,7 +120,7 @@ main() {
   local override_branch=        # Remote specified with -b
 
   local edit_wanted=false       # Edit commit message using -e
-  local commit_message=         # Custom commit message using -m
+  local wanted_commit_message=  # Custom commit message using -m
 
   local FAIL=true               # Flag for RUN: fail on error
   local OUT=false               # Flag for RUN: put output in $output
@@ -750,7 +750,12 @@ subrepo:commit() {
   o "Put info into '$subdir/.gitrepo' file."
   update-gitrepo-file
 
-  [[ -n $commit_message ]] || commit_message="$(get-commit-message)"
+  local commit_message
+  if [[ -n "$wanted_commit_message" ]]; then
+    commit_message="$wanted_commit_message"
+  else
+    commit_message="$(get-commit-message)"
+  fi
 
   local edit_flag=
   $edit_wanted && edit_flag=--edit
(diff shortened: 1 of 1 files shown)


ingydotnet:
@jrosdahl Any chance you can rebase this up to current master? I'm in the
middle of applying all outstanding PRs but too many conflicts here. I can try
again later in the week if you don't have time.

Thanks, Ingy

jrosdahl:
@ingydotnet: The fix is against issue/142, not master (where the bug doesn't exist). It looks like it's possible to apply it cleanly to issue/216 as well.

ingydotnet:
@jrosdahl, I cherry-picked your commit onto master, but there were 3 conflicts,
which I resolved, and pushed the result to issue/234 branch. Can you please
test and report back?

After that I'll rebase onto master and issue/216-B (the new dev branch).

jrosdahl:
Sure! I've had a quick look now.

The [rebased fix](612741950bb15d4fb3dce9014452616843b104e2) on issue/234

1. has a commit message that claims to fix a problem that doesn't exist in the commit's ancestry,
2. introduces an `edit_wanted` variable that isn't used for anything except to set an unused `edit_flag` variable and
3. introduces an undocumented `-N` option which doesn't do anything.

I don't think that it makes sense to try to apply the fix to master since it fixes a problem that isn't on master. You'll only get new rebase conflicts if you later try to re-rebase issue/234 to issue/216-B. Instead, I think you'd be better off just applying the fix directly to issue/216-B since the fix is for @grimmySwe's work done on issue/142. Or you could squash the fix into 9cd4ce7f85d454801291bb47a3dff7979834442f (previously known as c795fe1be827159d9c82da5224df36de3cff3df1; the bug was originally introduced in 865aa2e4b12931f897cd3a864c5a257c6561dfcb, I believe) if you like.

ingydotnet:
@jrosdahl, I get conflicts when rebasing this with issue/216 or issue/216-B.
Can you resolve these fr me?

jrosdahl:
This applies cleanly for me (with `issue/216-B` pointing to c01cf67):

```
git checkout issue/216-B
git cherry-pick fd31cf7
```

ingydotnet:
Aplied. Pushed. Thanks!

cherry-pick worked fine. :)

PR #237, opened 2016-11-02
Add GIT_SUBREPO_ROOT to manual installation instructions

### LINKED ISSUE #236: Incomplete instructions for manual installation
Because I'm on Windows and using [posh-git](https://github.com/dahlbyk/posh-git), I had to use the manual installation instructions, but I kept getting an error about "bash+.bash". Following the same instructions in bash (still on Windows) gave the same result. Apparently it's not enough to set the PATH environment variable, GIT_SUBREPO_ROOT has to be set as well.

thomaslevesque:
Fixes #236

### DIFF
--- a/ReadMe.pod
+++ b/ReadMe.pod
@@ -456,6 +456,7 @@ The second method is to do these things by hand. This might afford you more
 control of your shell environment. Simply add the C<lib> and C<man>
 directories to your C<PATH> and C<MANPATH>:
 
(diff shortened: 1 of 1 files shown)


ingydotnet:
Applied. Pushed. Thanks.

Actually, ReadMe.pod is a generated file. I took your fix and applied it to
`doc/git-subrpo.swim` which generates other files, and ran the update. Anyway,
it's all good now. Thanks.

Issue #220, opened 2016-09-01
`git subrepo status` and `git subrepo clean` autocompletion does not work in zsh


callalilychen:
Hi,

I have installed git subrepo and configured it for zsh. I tried to use subrepo with some test repositories. But `git subrepo status` and `git subrepo clean` autocompletion to add a subdir does not work and prints the following error message:

PAGER;col: -c: line 0: unexpected EOF while looking for matching `"'
PAGER;col: -c: line 1: syntax error: unexpected end of file

Under [http://pastebin.com/UwCZcMHn](url), you can find a log with `git subrepo status -x`. Can you help me to fix the problem?

Best regards,

Wenwen

perlpunk:
So the pasted output is from `git subrepo status -x`? That means, the error message is independent from completion.
I wonder where that line with

    /bin/sh -c '"unset' 'PAGER;col' ...

 comes from. Can you post your relevant `env`?

callalilychen:
Hi,

thanks very much for the quick response, I almost forgot it.

I saw the following item in my `env` output:
`
PAGER=/bin/sh -c "unset PAGER;col -b -x |     vim -R -c 'set ft=man nomod nolist' -c 'map q :q<CR>'     -c 'map <SPACE> <C-D>' -c 'map b <C-U>'     -c 'nmap K :Man <C-R>=expand(\"<cword>\")<CR><CR>' -"`

This is a environment variable, to config the output pager for `man`. Does subrepo also read the environment variable "PAGER"?

perlpunk:
I see. Yes, it's using PAGER.
Until someone figures out how to fix that, a quick workaround should be:

    export GIT_SUBREPO_PAGER=less

callalilychen:
thanks very much for the answer.

I just was in the mood to fix the problem und created a pull request :package:

ingydotnet:
@callalilychen, imho a much better solution here is to do something like this:

```
mypager() {
    /bin/sh -c "unset PAGER;col -b -x |
        vim -R
            -c 'set ft=man nomod #  nolist'
            -c 'map q :q<CR>'
            -c 'map <SPACE> <C-D>'
            -c 'map b <C-U>'
            -c 'nmap K :Man <C-R>=expand(\"<cword>\")<CR><CR>' -"
}
export -f mypager
export PAGER=mypager
```

That should work for all programs that honor `$PAGER`, without needing them to
`eval`, which is asking for trouble.

PR #238, opened 2016-11-16
fix pipe status to $PAGER evaluation problem

callalilychen:
see issue 220


ingydotnet:
@callalilychen, can you elaborate on which values of GIT_SUBREPO_PAGER that
this provides a fix for?

Using eval should be avoided unless absolutely necessary, and I can't figure
out what it would help with here.

ingydotnet:
This is better solved by putting a complex paging command into a function.
See example in #220 comment.

Closing.

PR #243, opened 2017-01-08
Fix a typo in section "git subrepo pull"

ebouchut:
Fix a typo in the section `git subrepo pull` where there is a space in the `update` command line option, in between the double dash and `update`.


perlpunk:
Thanks! The Readme.pod is generated from doc/git-subrepo.swim though, so Swim needs to be fixed here I guess. I'll create an issue there.

ebouchut:
Thank you Tina @perlpunk

PR #252, opened 2017-02-07
Fix typos

chetan51:
Replace `subrepo` with `subtree` in `subtree` example


admorgan:
Sorry about the false hope, did the wrong thing and it is getting fixed now.

PR #301, opened 2017-09-29
Ask git to find the .gitrepo files

jamezpolley:
Using `find .` can (a) take a long time and (b) go outside the current
repo.

Using `git ls-files` is inordinately faster, and only shows subrepos
of the current repo.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -1504,15 +1504,7 @@ assert-subdir-empty() {
 # Find all the current subrepos by looking for all the subdirectories that
 # contain a `.gitrepo` file.
 get-all-subrepos() {
-  local paths=();
-  paths=($(
-    git ls-files |
-      grep '.gitrepo$' |
-      grep -v '/.git/' |
-      grep '/.gitrepo$' |
-      sed 's/.gitrepo$//' |
-      sort
-  ))
+  local paths=($(git ls-files | sed -n 's!/\.gitrepo$!!p' | sort))
   subrepos=()
   local path
   for path in "${paths[@]}"; do
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -1504,8 +1504,10 @@ assert-subdir-empty() {
 # Find all the current subrepos by looking for all the subdirectories that
 # contain a `.gitrepo` file.
 get-all-subrepos() {
-  local paths=($(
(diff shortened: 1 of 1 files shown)


lib/git-subrepo
```diff
@@ -1458,7 +1458,8 @@ assert-subdir-empty() {
 get-all-subrepos() {
   local paths=();
   paths=($(
-    find . -name '.gitrepo' |
+    git ls-files |
+      grep '.gitrepo$' |
       grep -v '/.git/' |
```
chaoflow: @jamezpolley ``grep -v '/.git/'`` should not be needed when switching to ``git ls-files``

lib/git-subrepo
```diff
@@ -1458,7 +1458,8 @@ assert-subdir-empty() {
 get-all-subrepos() {
   local paths=();
   paths=($(
-    find . -name '.gitrepo' |
+    git ls-files |
+      grep '.gitrepo$' |
```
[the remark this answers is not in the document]
jamezpolley: Thanks! I've changed do doing this :)

lib/git-subrepo
jamezpolley: Thanks!

jamezpolley:
Rebased on top of 0.4.0

jamezpolley:
That test failure is weird. Running locally, I get a different test failure:

```
test/encode.t .................... 90/? #   Failed test 'subrepo push command output is correct'
#   at test/encode.t line 63.
--- /tmp/want-9606	2019-04-09 10:00:53.030626589 +1000
+++ /tmp/got-9606	2019-04-09 10:00:53.030626589 +1000
@@ -1 +1 @@
-Subrepo '~/......s:a^t?r a*n[g@{e.lock' pushed to '../../../tmp/upstream/bar' (master).
+Subrepo '~/......s:a^t?r a*n[g@{e.lock' has no new commits to push.
# Tests were run but no plan was declared.
test/encode.t .................... Failed 1/90 subtests 

```

but that test seems to have passed on travis, while the one that fails on travis passes for me locally.

jamezpolley:
It looks as though the travis issue is probably fixed in #358, so I'm keeping an eye on that. If that doesn't fix the problem I'll look into this more.

admorgan:
Merged to master

PR #376, opened 2018-10-27
Move the .gitrepo config file to .gitrepo/config

radix-platform:
I think it would be good to have .gitrepo hidden directory instead of file.


radix-platform:
Added fix for tests setup.

PR #379, opened 2018-10-27
Move the .gitrepo config file to .gitrepo/config (version 0.4.1 for release/0.4.0 branch)

radix-platform:
I think it would be good to have .gitrepo hidden directory instead of file.
I have fixed all tests and and checked the new functionality.


jrosdahl:
> I think it would be good to have .gitrepo hidden directory instead of file.

I think you forgot a crucial part of your pull request: describing *why* you think that it would be good.

radix-platform:
> > > I think it would be good to have .gitrepo hidden directory instead of file.
> > 
> > 
> > I think you forgot a crucial part of your pull request: describing _why_ you think that it would be good.
> 
> Because .gitrepo/config is similar to .git/config also hidden direcrories listed before hidden files and we have less chances of not seeing these files among our own code.

Also in the .gitrepo directories we can create tmp/subrepo folders for code (temporary saved) before create .git/reds/subrepo/... files

radix-platform:
Sorry, I did not tell about my suggestion in detail:
1) .gitrepo/config more clean for users because they read information from this file using git subrepo conig
2) in the future we will can to add more files in the .gitrepo directory (description at least)
3) we can use .gitrepo/tmp girectory in the same way as now we use .git/tmp in the corner of container repository to keep the .git directory clean
3) .gitrepo directory listed by `ls -Al' before then hidden file

By the way also I suggest to printout all content of .gitrepo/config file when the user call:

grimmySwe:
It sounds ok, was there a specific problem that you solved with this?

OleksiyRudenko:
@radix-platform suggestion sounds good.

What will (or will not) happen if I have already all things set up and then upgrade `git-subrepo` scripts? Is there a migration path for repos set up based on current `git-subrepo` implementation if any would be required?

radix-platform:
> It sounds ok, was there a specific problem that you solved with this?

I didn't faced any problem, but I thought that someday I would lose .gitrepo files among my own files ;)

radix-platform:
> @radix-platform suggestion sounds good.
> 
> What will (or will not) happen if I have all things already set up in some repo and then upgrade `git-subrepo` scripts? Is there a migration path for repos set up based on current `git-subrepo` implementation if any would be required?

I think you will not have problems with that because .gitrepo and .gitrepo/config on the version control by git and after update git-subrepo tool, when you will do git subrepo pull/push/clone -f the git-subrepo script will removes .gitrepo file and commit new .gitrepo/config .

radix-platform:
> > @radix-platform suggestion sounds good.
> > What will (or will not) happen if I have all things already set up in some repo and then upgrade `git-subrepo` scripts? Is there a migration path for repos set up based on current `git-subrepo` implementation if any would be required?
> 
> I think you will not have problems with that because .gitrepo and .gitrepo/config on the version control by git and after update git-subrepo tool, when you will do git subrepo pull/push/clone -f the git-subrepo script will removes .gitrepo file and commit new .gitrepo/config . Any case we have to double check this or make some tests.

I habe tested this situation on the release/0.4.0!

After update git-subrepo (with new place of .gitrepo) the command
$ git subrepo ../../remote/some.git subdir -b master -f 
doesn't work and say that subdir is not emty.
solution is that:

```
$ cs subdir
$ git mv .gitrepo config
$ mkdir .gitrepo
$ git mv config .gitrepo
$ git add .gitrepo
$ git commit -a -m "swiching to use new git-subrepo where .gitrepo file moved to .gitrepo/config"
$ git push

```
then this issue is fixed and you can continue to use new version of git-subrepo

```
$ git subrepo pull subdir/
Already up to date.

```

OleksiyRudenko:
@radix-platform seems like migration instructions and/or migration script are worth adding to the PR and docs.

It may become painful when a number of repos with multiple subrepos embedded would require manual or undocumented operations.

What do you think?

radix-platform:
> @radix-platform seems like migration instructions and/or migration script are worth adding to the PR and docs.
> 
> It may become painful when a number of repos with multiple subrepos embedded would require manual or undocumented operations.
> 
> What do you think?

Of course!
We have to create migration instruction which started from something like that:

Before update git-repo version please check all of subrepo
$ git subrepo status --all
...

Also we have to understand why `git subrepo clone url subdir -b master --force` doesn't work after update.

radix-platform:
After update git-subrepo on your machine we can update our container-repos by following procedure:
```
#!/bin/sh

subrepos="a b c"

for subrepo in $subrepos ; do
  ( cd $subrepo
    git mv .gitrepo .gitrepo.orig
    mkdir .gitrepo
    git mv .gitrepo.orig .gitrepo/config
    git add .gitrepo
  )
done

git commit -a -m "all subrepo .gitrepo config files are ready for git-subrepo-0.4.1"
git push

for subrepo in $subrepos ; do
  git subrepo pull $subrepo
done
```
Unfortunately output of `git subrepo status -q --all` command is not convinient foe a while named `for` like output from `ls` in Linux and we cannot say like follow:

```
for subrepo in `git subrepo status -q --all` ; do
  . . .
done
```

But I hope there is another way to list all subrepos directories related to the root of main repository.

jrosdahl:
@radix-platform: Thanks for the clarification!

> .gitrepo/config more clean for users because they read information from this file using git subrepo conig

Sorry, I don't understand your point here. The user can use `git subrepo config <subdir> <option>` regardless of the config file name, right? So why is `.gitrepo/config` cleaner when it comes to the `git subrepo config` command?

> in the future we will can to add more files in the .gitrepo directory (description at least)

That's true and could potentially be useful in the future.

> we can use .gitrepo/tmp girectory in the same way as now we use .git/tmp in the corner of container repository to keep the .git directory clean

That does not sound like a good idea to me. If `.gitrepo/tmp` is used for temporary files, then those files will show up in `git status`, etc, forcing the user to add `.gitignore` patterns. `.git/tmp` doesn't need that since `.git` is already handled specially.

> .gitrepo directory listed by `ls -Al' before then hidden file

(That's not the case for me, since I don't use the `--group-directories-first` option; I didn't even know it existed.)

>  and the user will not lose this file randomly.

Not sure what you mean by this. To my knowledge, I have never lost a file randomly despite not using  `--group-directories-first`. :slightly_smiling_face:

> > What will (or will not) happen if I have all things already set up in some repo and then upgrade git-subrepo scripts? [...]

> I think you will not have problems with that because .gitrepo and .gitrepo/config on the version control by git and after update git-subrepo tool, when you will do git subrepo pull/push/clone -f the git-subrepo script will removes .gitrepo file and commit new .gitrepo/config [...]

It's not that easy since git-subrepo reads `.gitrepo` files from commits in the ancestry to construct the proper commit tree before pulling/pushing from/to the upstream subrepo, so just renaming the files and committing won't be enough. If you don't provide backward compatibility, I think that current users will be annoyed (I certainly will be).

To sum it up, I don't think that it's worth the hassle to make this change unless there is a more concrete advantage than "could be useful in the future".

radix-platform:
> To sum it up, I don't think that it's worth the hassle to make this change unless there is a more concrete advantage than "could be useful in the future".

It is up to you to decide!

jrosdahl:
> It is up to you to decide!

Hm, it's not really for me to decide; I'm just a git-subrepo user with an opinion. :slightly_smiling_face:

ingydotnet:
Can you target this PR against master? The `release/0.4.0` branch has now been merged.

radix-platform:
> Can you target this PR against master? The `release/0.4.0` branch has now been merged.

Of cource. I will do it on my forked repo and try to pass all tests.

radix-platform:
I have created new PS and this PR can be removed.

PR #385, opened 2018-11-28
Make “git subrepo clean -f ...” delete refs correctly

jrosdahl:
Since Git can store refs in packed form (see git-pack-refs(1)), use `git update-ref -d` to delete refs instead of assuming that they are stored in the `.git/refs` directory.


admorgan:
Merged to master

PR #388, opened 2018-12-12
Make reclone with implicit branch update branch in .gitrepo correctly

jrosdahl:
If the subrepo is tracking a branch and you reclone using

    git subrepo clone --force $URL

without specifying -b/--branch then the branch setting in .gitrepo is
not updated.


admorgan:
@ingydotnet I would like your opinion on this one. While I agree it makes sense to follow the new branch, I am not sure we should force it.

PR #394, opened 2019-02-13
doc(pull): fix sentence repetition

jersou:
In "git subrepo pull" section, remove one of the :

> When you pull you can assume a fast-forward strategy (default) or you can specify a --rebase, --merge or --force strategy. The latter is the same as a clone --force operation, using the current remote and branch.
> 
> When you pull you can assume a fast-forward strategy (default) or you can specify a --rebase, --merge or --force strategy. The latter is the same as a clone --force operation, using the current remote and branch.


chaoflow:
The failing test seems not to be related to the change. The change here looks good to me.

admorgan:
Sorry about the false hope, did the wrong thing and it is getting fixed now.

PR #314, opened 2017-10-31
Fix #313: Ensured "-u" options is respected by "subrepo:push()".

plach79:
This is supposed to fix https://github.com/ingydotnet/git-subrepo/issues/313.

Issue #313, opened 2017-10-31
[release/0.4.0] "git subrepo push" always applies the -u option?


plach79:
I am not sure I found a bug or whether this is intended, but while on the `release/0.4.0` branch (https://github.com/ingydotnet/git-subrepo/commit/ba19aa75d3ceef126b1aa5ecfb8c66f8008fcd76), if I try to push while specifying an alternate remote and branch (I'm trying to implement a pull-request workflow for the subrepo), `.gitrepo` is updated with those remote and branch values, as if I specified the `-u` option.

This is not what I would expect, in fact I'd assume that, unless `-u` is specified, the original remote and branch are kept. Am I missing anything?

plach79:
Additional info: if I manually restore the original remote and branch in the `.gitrepo` file everything works as expected and I am able to pull from the (canonical) subrepo, after merging the changes pushed to my topic branch in my subrepo fork.

plach79:
PR with a possible fix for this at https://github.com/ingydotnet/git-subrepo/pull/314

grimmySwe:
I am thinking a little about what the expected behavior should be. Right now it seems that the .gitrepo file is always updated with new commit info, if you want to update the branch/remote you need to add the -u option.

For my own workflow, I would say that I prefer having the .gitrepo file update at all times, maybe introducing a flag that prevents update. So if I want to push stuff to another branch/repo nothing will happen locally.

What do you think about that?

plach79:
I may be wrong, I'm definitely new to `git subrepo`, but I tried `0.3.1` first and it seems to me that in that case the `-u` option was actually needed to update the `.gitrepo` info. This is also what I would expect from reading the (docs)[https://github.com/ingydotnet/git-subrepo/blob/master/ReadMe.pod#command-options]:

> If `-b` or `-r` are used, and the command updates the `.gitrepo` file, include these values to the update.

If I am correct, `0.4.0` would be introducing a BC break wrt `0.3.1` and adding a new flag to skip the update would work but would introduce an inconsistency with the rest of the commands.

Unless, obviously, the docs need to be updated and all commands are actually working the same way in `0.4.0` (didn't try that), that is always acting as `-u` was specified. This might be totally fine and a new flag to skip the update would make sense in this case, although it would still be a behavior change wrt the previous minor. 

[...]

PR #401, opened 2019-04-08
Ensure option -u is respected by subrepo:push()

aya:
rebasing @plach79 #314 PR onto master


admorgan:
@aya Do you have use case of this? If so could you comment on https://github.com/ingydotnet/git-subrepo/issues/313?

admorgan:
Closing as duplicate since this is a rebase and we can pull from the original.

PR #407, opened 2019-05-24
Make subrepo work when run in a worktree

### LINKED ISSUE #361: Not working with `git worktree` repo
```
git subrepo pull packages/public
git-subrepo: Command failed: 'git worktree add .git/tmp/subrepo/packages/public subrepo/packages/public'.
fatal: could not create leading directories of '.git/tmp/subrepo/packages/public/.git': Not a directory
```

This is because when you create a `git worktree`, `.git` is a file containing a link to the primary gitdir.

erijo:
In a worktree (created with git worktree add) there is no .git directory. It is
instead a file. To make subrepo work correctly in that case, use the
--git-common-dir argument to git rev-parse to find the path to the .git
directory. In a regular directory it just prints .git, but in a worktree it
prints the full path to the .git directory.

Fix #361

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -28,7 +28,7 @@ bash+:import :std can
 
 VERSION=0.4.0
 REQUIRED_GIT_VERSION=2.7.0
-GIT_TMP=.git/tmp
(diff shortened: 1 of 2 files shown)


admorgan:
Merged to master

PR #413, opened 2019-07-09
Explain required wd and branch for `subrepo clone`

jmcgeheeiv:
In particular, explain the message:
```git-subrepo: Can't clone subrepo. Unstaged changes.```
when the parent repository branch is not up to date with respect to the branch of the subrepo you are cloning.

### DIFF
--- a/ReadMe.pod
+++ b/ReadMe.pod
@@ -137,7 +137,13 @@ Add a repository as a subrepo in a subdir of your repository.
 
 This is similar in feel to C<git clone>. You just specify the remote repo url,
 and optionally a sub-directory and/or branch name. The repo will be fetched
-and merged into the subdir.
+and merged into the subdir. When you clone a subrepo:
+
+=item * The current working directory must be at the top of the parent repository.
+
+=item * The parent repository branch must be up to date with respect to the branch
+        of the subrepo you are cloning. The easiest way to do this is to set both
+        the parent repository and subrepo to the same branch (default C<master>).
 
 The subrepo history is I<squashed> into a single commit that contains the
 reference information. This information is also stored in a special file

### CODE REVIEW

File: ReadMe.pod [hunk 1]
```diff
@@ -137,7 +137,13 @@ Add a repository as a subrepo in a subdir of your repository.
 
 This is similar in feel to C<git clone>. You just specify the remote repo url,
 and optionally a sub-directory and/or branch name. The repo will be fetched
-and merged into the subdir.
+and merged into the subdir. When you clone a subrepo:
+
+=item * The current working directory must be at the top of the parent repository.
```
admorgan: I believe this would be better stated that git subrepo command must be run from the root of the git repo.

ReadMe.pod
```diff
@@ -137,7 +137,13 @@ Add a repository as a subrepo in a subdir of your repository.
...
+
+=item * The parent repository branch must be up to date with respect to the branch
+        of the subrepo you are cloning. The easiest way to do this is to set both
+        the parent repository and subrepo to the same branch (default C<master>).
```
admorgan: I have been trying to parse this, and even having a firm understanding of subrepo I can't grok this. I read this as clone can't import anything except the newest change on the branch. I don't understand what you are trying to impart in "setting both the parent repository and subrepo to the same branch."

ReadMe.pod
jmcgeheeiv: I agree.

ReadMe.pod
jmcgeheeiv: I think I will retract this as it is very difficult to document the general rule. A better approach would be to improve the error message (which I do not recall). The error message would have the concrete situation at hand, making it easier to explain, and the user need not have read the documentation.

jmcgeheeiv:
I will retract this because the general situation is too hard to explain. When I get the chance, I might fix the error message.

PR #415, opened 2019-07-19
Clear error when rebase has broken parent reference

OskiKervinen-MF:
I had rebased the commit where I performed the `git subrepo` clone and when I tried to do a `pull`after that, I received the error
```
git-subrepo: Command failed: 'git branch subrepo/utilities/cacherer '.
fatal: Not a valid object name: ''.
```
I was able to figure out this was because of the rebase eventually, but it took some doing, so I made this MR to make the problem obvious in the future.

Also replaced one raw reference to .git with an invocation of `git rev-parse --git-dir`.


admorgan:
I like both of these changes

OskiKervinen-MF:
I removed the part that touched the submodule issue. The remaining test failures seem to equal those in master.

arichardson:
I've accidentally rebased after doing a subrepo push/pull a few times and then getting weird failures the next time I try to update. Having a meaningful error message would be really useful.
Thanks!

PR #419, opened 2019-08-13
Fix parent update

serbogus:
I had found that the parent pointer does not change after the pull. As a result, the next pull does re-sampling of old commits and ends up in merge conflicts in some cases. To fix the issue, i commented out the extra check in the update-gitrepo-file (# if [[ $upstream_head_commit == $output ]]; then). 

This works in all my cases. It also works in all tests but "pull-merge.t".  The test did sample the parent before pull. So, i modified it to sample the parent after the pull.

There was also an issue with 'push.t'. The push after pull had truncated history, because the parent pointer had been moved before the push is done. The solution for the issue was to add yet another parent pointer as 'push-parent'. So, the .gitrepo has another entry now. 

On the way i had fixed a few issues:

1. git 2.21 changed word casing in the 'Cc]ouldn't find remote ref' message. I fixed  the regex

1. branch... tests were failing sporadically at least with git 2.22. The order of transactions reported by rev-list in presence of branches was different from test to test. looks like --topo-order fixed it.

1.  error.t was failing inside a submodule, because there was a ref to '.dir'.  So, the 'cd' command failed. I fixed the test.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -114,6 +114,7 @@ main() {
   local subrepo_branch=         # Upstream branch to clone/push/pull
   local subrepo_commit=         # Upstream HEAD from previous clone/pull
   local subrepo_parent=         # Local commit from before previous clone/pull
+  local subrepo_push_parent=    # Local commit from before previous push
   local subrepo_former=         # A retired gitrepo key that might still exist
 
   local refs_subrepo_branch=    # A subrepo ref -> commit of branch/pull command
@@ -494,7 +495,7 @@ subrepo:clone() {
 
   o "Commit the new '$subdir/' content."
   subrepo_commit_ref="$upstream_head_commit"
-  CALL subrepo:commit
+  CALL subrepo:commit clone		
 }
 
 # Init a new subrepo from current repo:
@@ -604,10 +605,12 @@ subrepo:push() {
     if $squash_wanted; then
       o "Squash commits"
       subrepo_parent="HEAD^"
+		elif [ -n "$subrepo_push_parent" ]; then
+			subrepo_parent=$subrepo_push_parent
     fi
 
     o "Create subrepo branch '$branch_name'."
-    CALL subrepo:branch "$branch_name"
+    CALL subrepo:branch "$branch_name" 
     cd "$worktree";
 
     if [[ "$join_method" == "rebase" ]]; then
@@ -669,6 +672,8 @@ subrepo:push() {
   upstream_head_commit="$new_upstream_head_commit"
   subrepo_commit_ref="$upstream_head_commit"
   update-gitrepo-file
+	RUN git config --file="$gitrepo" subrepo.push-parent "$original_head_commit"
+	RUN git add  -f -- "$gitrepo"
   RUN git commit -m "$(get-commit-message)"
 }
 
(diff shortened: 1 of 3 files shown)


jakebailey:
I know this is an old PR (and appears to be inactive), but the parent check is affecting a subrepo I've got where the `parent` ref is never never updated until another re-clone (which is a shame). I'd love to see some version of this merged so I can depend on it.

admorgan:
The reason the parent SHA isn't being updated is that you have changes that are on the local version that haven't been pushed upstream. Therefore the only safe thing to happen is for us to replay the changes and try to integrate the pull into them. I have found that I get a lot more merge failures when using the rebase method than the merge method. Unfortunately that particular patch is unlikely to ever be merged. The section I have marked for 4.2 is patch 
1dd3c7d, the rest of them will be discarded.

jakebailey:
Your reply confuses me. I am pushing changes upstream (though not via `subrepo push`, and I'm using the merge method, not the rebase method.

I can't use `subrepo push` because:

1. It modifies the `.gitrepo` file to change the remote path if `--remote` is provided. I split out changes from my subrepo into a PR from a fork to push upstream. When I do this, `subrepo push` creates a commit on the repo containing the subrepo changing the metadata. I do not want to push changes directly upstream without vetting.
1. I want to squash the outgoing changes as the transferred commits are not clean at all. They include the `git subrepo` commits which I do not want to push upstream; subtree is an implementation detail of my repo, not the repo I've vendored.

Instead, I `git subrepo branch`, then squash in the worktree, then push the subtree branch to my fork of the external subrepo, where it gets pulled in. Then, the next time I `subrepo pull`, things should be getting synced, but only `commit` changes.

jakebailey:
If you think my problem is out of scope for this PR or wouldn't be fixed by it, then I can open a new issue. I only commented here because I thought this PR was open, could be merged, and might have fixed my issue.

admorgan:
This PR would work for your specific workflow, but it breaks a number of others. I would prefer you open another ticket so it would be easier to find the discussion. Unfortunately there isn't currently a great way to handle your issue and I think it is going to require some discussion.

jakebailey:
Sure, I'll open another issue with my usecase.

I did think up a workaround for now; run `git subrepo branch` and check if there are any differences between `HEAD` and `refs/subrepo/<name>/fetch`. If there are none, then it's safe to reclone (or maybe just change `parent` directly).

thorsten-klein:
Hello @jakebailey,
+1
Do you have the exact commands you run? :-)

jakebailey:
Sure, it's roughly this:

```sh
$ git subrepo clean <name>
# Split out the changes still applied to the subrepo but not upstreamed.
$ git subrepo branch <name>
$ pushd .git/tmp/subrepo/<name>
# Diff the worktree with upstream.
$ git diff --quiet <new hash of subrepo>
# If this succeeds, then:
$ popd
$ new_parent=$(git rev-parse HEAD)
$ git config --file path/to/.gitrepo subrepo.parent $new_parent
# Shift up the parent commit to simulate a forced reclone.
$ git commit -am "Update git-subrepo parent"
```

But, in a script. Alternatively if the previous commit was a `git subrepo pull` (what we do), the last step could `--amend`, or just roll back the last commit and reclone again as we know it's safe.

PR #420, opened 2019-08-21
Eliminate the need for GIT_SUBREPO_ROOT to be set in the zsh completion handler

### LINKED ISSUE #183: Remove requirement for $GIT_SUBREPO_ROOT from _git-subrepo zsh completions
When the `_git-subrepo` completion script is dropped directly into `/usr/share/zsh/site-functions` the completions JustDon'tWork(tm).

Furthermore, the check doesn't actually DO anything except bark and barf if the variable hasn't been set - the completions work just fine without it.

ryantrinkle:
Fixes #183

I don't know how to comprehensively test this, but I have verified that some basic completions work correctly


admorgan:
Was this code introducing an issue or just unnecessary?

ryantrinkle:
@admorgan The code was causing me trouble because I packaged git-subrepo for the [NixOS](https://github.com/NixOS/nixpkgs) linux distro, so I had it installed that way rather than via a local git repo checkout.  I also use zsh, so I wanted to have the completions working.

I couldn't find anywhere the zsh completion script actually depended on the local repo approach, so it seemed cleanest to remove the check.  However, if I've overloooked something, I'd be happy to find a better solution!

admorgan:
I asked so we could add a test to the travis to make sure things like this work correctly in the future.

ryantrinkle:
@admorgan Ah, interesting.  Well, testing the exact situation I was in would look something like "install via a linux package manager, then make sure command completion works in zsh".  That seems like a somewhat challenging test to write (or at least challenging to make it fast!).

admorgan:
How to you get the package into the "path"? In reality it is simply reproducing this interchange that we need to test, not the install step.

mvz:
Wouldn't a reasonable test be unsetting GIT_SUBREPO_ROOT and confirming the completion still works?

ryantrinkle:
With Nix, when you install a package, anything in its `bin` directory is symlinked into a location that is on the user's path - in my case, `/home/ryan/.nix-profile/bin/git-subrepo`.  I don't know in much detail how this works with other package managers.

Perhaps the right test would be be to put git-subrepo in a weird location, put it on the path, and then check that completion still works?

admorgan:
I agree, this would mean that we need to rethink what we look for to ensure that completion is successfully plumbed in. The variable may not be the best way any longer.

Issue #389, opened 2018-12-19
Performance with 1000+ subrepos


udeved:
Hello,

this is not directly a bug report or feature request, rather than asking for advice.
We use subrepos for our package git tree, one subrepo package(PKGBUILD) for each.

This amounts to roughly 1100 subrepo in one main project.

Now the problem, a push, usually the first push after some subrepo has been cloned, depending on the commit count, tend to be painfully slow when subrepo rewrites history. If I enable -v switch, I see subrepo sequentially interating throught the commits.
In a repo with about 4000 commits and 1100 subrepos, it takes up to 4-5 mins to push a change into a subrepo.

My question is, if there is something that can be done abut that, or if I might use subrepo command wrongly? 

We use a simple `git subrepo push $subrepo --clean`. Is the clean flag responsible for the slow initial pushes? Documentation is somewhat sparse on that.

grimmySwe:
Hi @udeved, yes, there are some operations that can take some time with subrepo. Are you using the latest master? or the "old" 0.3.0 release?

number of subrepos shouldn't be a problem in itself as you only "work" against one at the time, although if you have that many subrepos there will be a lot of commits just handling them in the main repo.

udeved:
@grimmySwe 

We are still stuck with https://github.com/chrisfair/git-subrepo fork, since it was already 0.4.1 long before 0.4.0 was relased. I may give it a try from upstream master.

Problem is, I need to mass update all Jenkinsfile in the subrepos, and many subrepos had no commit since clone.

udeved:
another question on the --all option.

Is it viable to use a git subrepo push --all?
Is it faster that pushing each single one? Ie, does the reqrite with --all map all subrepos, of for each just as normal push does?

udeved:
Regarding commit handling, our superrepos holding the subrepos are basically irrelevant. They are just a nice means to get all the package repos in one go. Our build operations etc are triggered by pushes in the subrepos, so a simple git prune in the superrepo does the job for us after each push.

grimmySwe:
Ok, sounds like a cool use case. Nothing we have prepared for though.

Regarding your "all" question, there already is a an issue for this in #318.
Your "None" suggestions sounds interesting, it could certainly be helpful when you handle lots of subrepos. We'll see what 2019 brings :-)

udeved:
@grimmySwe 

Thank you for your answers, its appreciated. :)

[...]

PR #436, opened 2020-01-21
Use git-filter-branch --index-filter instead of --tree-filter

### LINKED ISSUE #435: Should not use filter-branch --tree-filter
I have been using git-subrepo in various projects and the performance doesn't really matter. However, I recently updated a mirror of the libc++ subdirectory using git-subrepo and after an updat, wanted to pushed about 10000 new commits (of which only a fraction touch the subdirectory).

git-subrepo spends many hours running:
`it filter-branch -f --prune-empty --tree-filter rm -f .gitrepo db47e4cfa09972aa929ca352b8bfdc64a634d855..subrepo/libcxx`

Using a tree-filter is really slow because git has to perform a checkout for every commit.
Ideally git-subrepo should be detecting the presence of the faster [git-filter-repo](https://github.com/newren/git-filter-repo/) and using that instead.

But even with filter-branch it should be possible to get much better performance by using an index-filter:

I haven't tried this yet, but looking at the examples section of the filter-branch documentation (https://git-scm.com/docs/git-filter-branch#_examples), the following might work:

`git filter-branch --index-filter 'git rm --cached --ignore-unmatch .gitrepo' db47e4cfa09972aa929ca352b8bfdc64a634d855..subrepo/libcxx`

arichardson:
Using an index filter is much faster since git does not have to perform
a full checkout for every commit in the range that needs to be rewritten.

Fixes #435

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -816,8 +816,8 @@ subrepo:branch() {
   o "Remove the .gitrepo file from $first_gitrepo_commit..$branch"
   local filter=$branch
   [[ $first_gitrepo_commit ]] && filter=$first_gitrepo_commit..$branch
-  FAIL=false RUN git filter-branch -f --prune-empty --tree-filter \
-    "rm -f .gitrepo" "$filter"
+  FAIL=false RUN git filter-branch -f --prune-empty --index-filter \
+    'git rm --cached --ignore-unmatch .gitrepo' "$filter"
 
   git:create-worktree "$branch"
 


udeved:
@admorgan 
 
Wrt

> 
> use_tree_filter subrepo branch normally uses 'index-filter' for performance reasons.
> 						It is noted that this could hit some git limitations for huge trees and many revisions.

Do you happen to know specific limitations or possible errors?

I had opened an issue regarding performance with many subrepos here.
https://github.com/ingydotnet/git-subrepo/issues/389

By now, we reduced the master repo and split it up into smaller units, but our biggest master repo still has about 600 subrepos.

I guess I am asking, what could happen worst case if we tried the faster filer option? Its kind of risky and but unclear in the comment.

admorgan:
I have been trying to figure out a way to have a fast/slow path for this one without success so far. "...what could happen worst case if we tried the faster filter option" is you run out of RAM an you leave the repo in an inconsistent state. Before that happens your machine becomes completely unresponsive and you are cursing ever encountering subrepo.

The problem I am having is I don't know how to identify we will get in that state before hand, and I can't figure out a way to watch the process and kill it and start over with the slow method if it gets out of hand. I firmly believe at 99.9999999% of people would never see this event and your patch is absolutely the right way to go, but being in consumer electronics means that is still a really large number of failures. I have considered adding a switch to let you enable the fast path, and on my machine making it default.

I am sorry I don't have a better answer. As always I am happy to work through any ideas anyone else has.

arichardson:
@admorgan Do you have an example repository where this OOM situation occurs? 
Maybe it only happens with specific git versions so we could use a version check to enable index-filter? Or does it depend on the number of commits/files that are processed? I have not seen any issues with index-filter for repositories as large as https://github.com/llvm/llvm-project or https://github.com/freebsd/freebsd.

admorgan:
I created a completely fabricated one that was beyond reason, of course I have 64G of RAM, so it takes something that ridiculous to cause issues. The problem is not number of commits, but the size of them and the complexity of the diffs. Huge numbers of huge commits significantly modifying large numbers of huge files is the only way I could run into an issue. But when I did it was spectacular.

admorgan:
I have to be honest, neither of the example repos are anywhere close to the sizes I use on a daily basis at work.

udeved:
I have been in the mood of risk and tested the patch on our 600+ subrepo python.
It turns out, there is almost no noticeable improvement sadly, a one year old commit to push took still about 5mins. Two other team mebers also tested it with the same result.

In my case, the 1 year old commit took 5mins with about 9800 objects mapped in filter branch.

admorgan:
This information makes me very sad. I was really hoping this would get me out of a few long operations on some of my systems.

arichardson:
For me, the other step that takes a long time when I use `git subrepo push` it's the `Create new commits with parents into the subrepo fetch` step. This takes a really long time if I push many commits.

Example debug output (it seems like it's at least one `git config` plus one `git cat-file` command for each commit.
```
>>> git config --blob fae035daed20413633cafe1cb0d2eeb50e78b85d:libcxx/.gitrepo subrepo.commit
  * .gitrepo reference commit: 3ee695ba987133ff472e598f8176c28e170458d5
  * is child: a569157c7781a957f8a32a28575b03d0a7f226d5 dcc04e09cf6e36c6d2c47057a8f201f2c0784c69
  * Check for rebase
  * Find parents
  * Create a new commit -p 7ff0a2e66b8adc833037bbf3fe456d80201097fd
>>> git cat-file -e fae035daed20413633cafe1cb0d2eeb50e78b85d:libcxx
  * Create with content
  * Working on c96213389740090c26c0df10f6846cc494b9ed41
>>> git config --blob c96213389740090c26c0df10f6846cc494b9ed41:libcxx/.gitrepo subrepo.commit
  * .gitrepo reference commit: 3ee695ba987133ff472e598f8176c28e170458d5
  * is child: fae035daed20413633cafe1cb0d2eeb50e78b85d f757ecbf85605735195441abefd9c291f5e317cc
  * Check for rebase
  * Find parents
  * Create a new commit -p c67af1541aed948846ddfe2bc0cd5cd593405f1b
>>> git cat-file -e c96213389740090c26c0df10f6846cc494b9ed41:libcxx
  * Create with content
  * Working on e1b6230d27bfb61ad3e7094d86f1c46bbd9c7091
>>> git config --blob e1b6230d27bfb61ad3e7094d86f1c46bbd9c7091:libcxx/.gitrepo subrepo.commit
  * .gitrepo reference commit: 3ee695ba987133ff472e598f8176c28e170458d5
  * is child: c96213389740090c26c0df10f6846cc494b9ed41 94b195ff1236004a0b74a63803b71d45a8ca496a
  * Check for rebase
  * Find parents
  * Create a new commit -p a83af77bd890274a5dffba8aad25061ee3d9d11c
>>> git cat-file -e e1b6230d27bfb61ad3e7094d86f1c46bbd9c7091:libcxx
  * Create with content
  * Working on 48087b4859af1c6154a19a2828aa54c6b4cce611
>>> git config --blob 48087b4859af1c6154a19a2828aa54c6b4cce611:libcxx/.gitrepo subrepo.commit
  * .gitrepo reference commit: 3ee695ba987133ff472e598f8176c28e170458d5
  * is child: e1b6230d27bfb61ad3e7094d86f1c46bbd9c7091 767a97b22339350d1d1368ca5548f967e49a4c00
  * Check for rebase
  * Find parents
  * Create a new commit -p 4c98c548df058333b758cd4cf102cbb3f40a250c
>>> git cat-file -e 48087b4859af1c6154a19a2828aa54c6b4cce611:libcxx
  * Create with content
  * Working on 2846c3c091365abafeea4660a9c2d471cc805e0e
>>> git config --blob 2846c3c091365abafeea4660a9c2d471cc805e0e:libcxx/.gitrepo subrepo.commit
  * .gitrepo reference commit: 3ee695ba987133ff472e598f8176c28e170458d5
```

I guess this would be a lot faster if we weren't spawning new processes and instead using something in-process (e.g. pygit2?).

udeved:
The problem we have is, since pushes in large repos take so long, it can happen, by the time the push went out, the master repo is already outdated, despite a pull before the subrepo push.

@admorgan 

Makes me kind of sad too, I ws so hoping we could reintroduce one big master repo eventually, which would simplify our tools and workflow quite a bit. I still keep up hope though. :)

admorgan:
I may have some good news on that front. The push behavior does not work well with gerrit or PR workflows. Therefore I never use it directly. I am going to propose a new workflow to make those style interactions more first class citizens as I expect they are the common case.

udeved:
A little report after a while.
Our git-subrepo is now by default patched with this PR, ie all team members use it.

No worst case crash or git or so experienced so far, on that front, the PR works (for our purposes)
We don't do any PRs on the master repos, its plain git and sub repo push & pull.

It appears, in out setup, the mentioned problems start with amaster repo size of ~400 subrepos.
~200 subs is ok push times, at 400 it can get slow, and worst is above 600 subs, where a pretty recent commit of a sub can still take up to a minute (each sub has had a commit recently for python3.9 rebuilds).

admorgan:
Closing because still haven't found a workaround for the worst case.

PR #443, opened 2020-03-09
Fix error when sourcing .rc with old zsh version

jrosdahl:
zsh 5.0.2 has trouble parsing `[[ ${BASH_VERSION} ]]` (see 27f51ea8e7927) in the .rc file:

```bash
% zsh --version
zsh 5.0.2 (x86_64-redhat-linux-gnu)
% source git-subrepo/.rc
git-subrepo/.rc:16: parse error near `]]'
```

### DIFF
--- a/.rc
+++ b/.rc
@@ -13,7 +13,7 @@
 #
 #------------------------------------------------------------------------------
 
-if [[ ${BASH_VERSION} ]]; then
+if [[ -n ${BASH_VERSION-} ]]; then
   if [[ ${BASH_VERSINFO[0]} -lt 4 ]] ; then
     echo "The git-subrepo command requires that 'Bash 4+' is installed."
     echo "It doesn't need to be your shell, but it must be in your PATH."


.rc
```diff
@@ -13,7 +13,7 @@
 #
 #------------------------------------------------------------------------------
 
-if [[ ${BASH_VERSION} ]]; then
+if [[ -n ${BASH_VERSION-} ]]; then
```
admorgan: I see that ${BASH_VERSION-} works on my machine, but why is it necessary?

jrosdahl: > I see that ${BASH_VERSION-} works on my machine, but why is it necessary?

Don't know if it's necessary. It indeed looks like the `-` isn't needed for zsh 5.0.2, zsh 5.7.1, bash 4.2.46 and bash 5.0.11 at least. The reason I added the `-` was to make it analogous to https://github.com/ingydotnet/git-subrepo/blob/a04d8c2e55c31931d66b5c92ef6d4fe4c59e3226/.rc#L29 which seems to originate from dee5311a154d9544ddf85f9e233281c9dcbedb8a.

(The `-n` is however needed in zsh 5.0.2.)

Do you want me to remove the `-` in one or both places?

admorgan: Ok I believe I have found an answer. The minus seems to be short hand for :- in other words, it says that if the value is set, but is NULL then it will return an empty string. This prevents finding a value that we can't use. Therefore I agree with your change as is. If you rebase this to the release/0.4.2 branch I will accept it today.

jrosdahl: Yes, that's what it does but I still don't see why it (or `:-`) would be needed in this case since `[ -n $unset_variable ]` works in practice as well.

Anyway, rebased to release/0.4.2 now.

admorgan:
Sorry this took so long. I missed the e-mail some how.

PR #284, opened 2017-06-28
Added (very) basic fish support

ErikBjare:
Slight progress on https://github.com/ingydotnet/git-subrepo/issues/187.

Doesn't do autocompletion, but the other stuff works.

By the way, thanks @ingydotnet for working on this project! 

Hopeful it will be of use for my ActivityWatch project (see this issue: https://github.com/ActivityWatch/activitywatch/issues/70).

PR #446, opened 2020-03-18
Add a fish shell startup script

nuxeh:
This adds the subrepo command, but adding to the MANPATH variable in fish seems to cause manpages to not work for other things, from anecdotal evidence. Bash completion clearly doesn't work, but is something that could be done for fish shell at a later point, as fish shell completion definitions.

Since most of subrepo is executed explicitly with bash, it seems to work fine in fish shell as far as I can tell, besides the need for a fish rc file, as added in this commit.

### DIFF
--- a/.fish.rc
+++ b/.fish.rc
@@ -1,0 +1,16 @@
+#!fish
+
+#------------------------------------------------------------------------------
+#
+# This is the `git-subrepo` initialization script.
+#
+# This script turns on the `git-subrepo` Git subcommand for the *Fish* shell.
+#
+# Just add a line like this to your `~/.config/fish/config.fish`:
+#
+#   source /path/to/git-subrepo/.fish.rc
+#
+#------------------------------------------------------------------------------
+
+set GIT_SUBREPO_ROOT (readlink -f (dirname (status --current-filename)))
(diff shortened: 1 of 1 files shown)


admorgan:
I noticed there are some subtle differences between this and #284 can you look at them and update this one if there are any benefits to what they were using. I don't know fish, or have an ability to test it so I will have to take your word for the fact that it works.

admorgan:
Could you also rebase it on release/0.4.2?

nuxeh:
Sure, rebased. Here the fish rc script uses a slightly different commands to get the current directory the script is run in (i.e. the git-subrepo install directory), but are entirely equivalent as far as I can tell. I just followed, as far as I could, guidelines I could find on how to update PATH and MANPATH. Works for me, at least.

admorgan:
I am not sure what happened, but it looks like github doesn't think it is pointing at the correct branch.

nuxeh:
Yes, good point, just changed the base on Github.

nuxeh:
As to whether there is anything in the previous PR which would be beneficial, the only thing is the extra parameters to the `set` command, which doesn't seem to be required at least on recent fish on Linux. Also, the previous PR's method of updating MANPATH won't work correctly, as far as I know.

Issue #448, opened 2020-03-19
git subrepo clean no longer working in 0.4.1


rdzman:
I have two projects with their own repos (call them A and B), where B is a subrepo of A. I normally do all my commits for B in B's main repo, then from within A I do `git subrepo pull B`, followed by `git subrepo clean B`. The latter always got rid of the `subrepo/B` branch.

However, today I did the same thing after updating from git-subrepo 0.4.0 to 0.4.1 and the `subrepo/B` branch is still there.

Thinking it might be related to a0c967d, I tried `git show-ref` and it does not show any lines with `refs/subrepo/B`, but the directory `refs/subrepo/B` is still there.

Is this a bug?

rdzman:
I should have mentioned that this is on macOS Catalina 10.15.3, with the default zsh as my shell and bash 5.0.6 installed in /opt/local/bin by MacPorts.

admorgan:
This is a bug. Of course the real question is, is it a bug in subrepo or git. Now to find out why isn't the ref showing up in git show-ref.

rdzman:
If you can't reproduce it, let me know if there's something you'd like me to try on my end to track it down.

admorgan:
I probably won't be able to look at it till some time tomorrow. I will let you know what I discover. I know that we have a test in our CI to check for this, but I don't know what version of git our CI is using off the top of my head.

rdzman:
Any news on this yet?

rdzman:
This does seem to have been fixed by #455.

PR #455, opened 2020-04-16
Fix unreliable branch existence check

jonathn:
It can sometimes be the case that `.git/refs/heads/subrepo/mylib` is not present even though branch `subrepo/mylib` does actually exist.
This change fixes that and makes the code more consistent by re-using an existing function.


admorgan:
If you apply this to release/0.4.2 I will merge it immediately

admorgan:
This may address #448

admorgan:
Merged on release/0.4.2

PR #456, opened 2020-04-16
Don't suppress stderr when git merge fails

jonathn:
If `git merge` fails for whatever reason and prints the error on `stderr` the current implementation silently discards it, leaving the user with no idea what went wrong. There's no advantage in discarding `stderr` in this case.


admorgan:
If you apply this to release/0.4.2 I will merge it immediately

admorgan:
Merged to release/0.4.2

PR #475, opened 2020-06-03
git-fetch: Enable progress indicator [WIP]

icy:


admorgan:
Did you decide this wasn't viable?

icy:
> Did you decide this wasn't viable?

I tried to turn off some option, but that also requires to rewrite the tests which depend on the output of commands. I would think of another way, maybe you have a better idea.

icy:
I ended up in writing my own tool, to resolve my own problem: https://github.com/icy/git_xy#why . Hope that I have another chance to look at this issue and the project. 

Best

Issue #471, opened 2020-05-27
'git-subrepo' DEB packaging


spog:
Hi,

I made a sample project for packaging your great git-subrepo extension into a DEB package. The packaging project itself (https://github.com/spog/git-subrepo-debian.git) uses a subrepo to import required version of 'git-subrepo' code.
There is also an Ubintu-bionic 'git-subrepo' packaga already available on my PPA: https://launchpad.net/~spog/+archive/ubuntu/ppa

I hope there is some interest about packaged 'git-subrepo'?

regards, Samo

p.s.
I would suggest, that you mention somewhere in docs, that specifying a branch on the 'subrepo' command line may also be a tag name or a commit checksum. It is not clear, since i ran into a tutorial (https://sudonull.com/post/8961-Git-subrepo) explicitly stating: ...the user cannot connect a subproject referring to his fixed tag or a specific revision,...  or maybe i got it all wrong

admorgan:
I think it is awesome that you packaged it for Debian.

As for using an explicit rev/tag for the branch does work. It prevents the pull feature from working, but I use that feature extensively for some of my use cases. I agree this should be added as an advanced "feature" in the documentation.

spog:
Thanks for the quick reply.
There is a glitch in DEB auto-completion support, since (ubutu-18.04 / git-2.17 - my installation) does not provide dynamic completion scripts loading from the git completion script itself. The next git-2.18 however does. Currently i need to source the DEB installed '/usr/share/bash-completion/completions/git-subrepo' via .bashrc or to fix git's '/usr/share/bash-completion/completions/git'.
I'd like to hear, if there were some other issues too.

perlpunk:
Good idea! I'm working on getting this into openSUSE too
https://build.opensuse.org/package/show/home:tinita:branches:devel:tools:scm/git-subrepo

admorgan:
Just a heads up, but 2.17 is the oldest version of git that is supported. I don't even have a CI system to test it against any longer. I am really excited about all the packaging.

spog:
Thanks for more info. I am currently preparing packages for all live Ubuntu distros from 16.04 up to 20.04. and then think about Debian itself. Does it make sense to make package for Xenial (16.04  with git-2.7.4) - already done - as you mentioned 2.17?

perlpunk:
> 2.17 is the oldest version of git that is supported

Did you mean 2.7?

admorgan:
[...]

PR #476, opened 2020-06-03
zsh completion: Remove check for GIT_SUBREPO_ROOT

perlpunk:
I can't remember why I added this check.

This prevents zsh completion from working when installed as a rpm
package.

See also https://github.com/ingydotnet/git-subrepo/issues/471#issuecomment-638365294

I didn't regenerate the completion yet.


admorgan:
Can you put this on release/0.4.2 instead?

admorgan:
Still says it is on master

perlpunk:
Oh, I see, you meant the target branch. Fixed.

PR #479, opened 2020-06-06
provide activation script for Windows cmd

flashdagger:
`activate.bat` is the Windows cmd equivalent to `.rc` 
It first checks if the git command is available. Then sets environment variables `GIT_SUBREPO_ROOT` and `PATH`.
Multiple calls do not clutter up the `PATH` variable.


NexZhu:
@admorgan Can this be merged?

PR #489, opened 2020-07-25
Feature/reset

flashdagger:
This PR adds the 'git subrepo reset' subcommand.

It simply does a forced reclone on an existing subrepo. 
The remote and branch are taken from the .gitrepo file if not specified. You can optionally also specify a commit-ish reference.

Some people asked for something like this and I also sometimes need to set the subrepo to a certain state when something is wrong or I just need to switch back.
Since I'm not an  bash expert the code might not be perfect and I didn't provide any tests. But the scripts are well structured so I could easily implement it for myself. Would be happy to get some feedback.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -37,6 +37,7 @@ git subrepo <command> <arguments> <options>
 Commands:
   clone     Clone a remote repository into a local subdirectory
   init      Turn a current subdirectory into a subrepo
+  reset     Reset subrepo performing a forced reclone
   pull      Pull upstream changes to the subrepo
   push      Push local subrepo changes upstream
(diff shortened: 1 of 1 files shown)


jrosdahl:
Just a note: The described use case is partially related to #388.

flashdagger:
Yes. I think we all need something similar.
One thing I like about subrepo is that it keeps the information about the remote and branch so I don't need to specify it on every command like for git subtree. But once something is messed up or you decide to switch branches you need to use subrepo clone and specify everything from scratch...

admorgan:
You can now use --force with git pull as the documentation always said you could, but recently was added.

PR #492, opened 2020-08-09
run index-filter instead tree-filter to remove .gitrepo file

flashdagger:
the index-filter option might be faster, as it does not need to touch the working directory


admorgan:
Duplicate of #436

Issue #424, opened 2019-10-31
fatal: Not a valid object name: '' no2


forry:
Hi,
I'm not a git expert. I'm using subrepo for one of the modules (dormon/geGL) of our framework (Rendering-FIT/GPUEngine). I don't have push rights to this module. I have done one local change to the module and committed it as part of the framework repo. After while I wanted to pull new version getting:
```
git-subrepo: Command failed: 'git branch subrepo/geGL '.
fatal: Not a valid object name: ''.
```
I have read through issue #350 but I don't think it applies to me since I haven't changed the remote or the .gitrepo by hand. I have tried clean and doing the suggested commands by hand.

I just need to merge in new changes. Is there another way apart from deleting and recloning the subrepo from scratch. And how to avoid this issue in the future?

Thanks, forry

admorgan:
Is this something that you would still like help with? I know it has been a long time. I am sorry for the silence.

forry:
Hi,
it's ok, and yes I would very much like to get help :). At least I will learn something. As I remember it the problem was, that there were a couple of files with mixed tabs and breaks in the subrepo. Git was constantly showing these files as modified but then it said that there are no changes.  So I committed them as a local change and probably not via the subrepo command but to the parent repo. Then I was making some experiments and ending with the message above (don't remember exactly, a lot was going on and I wasn't working on that project since). And ofc I don't have a clue of how to make it right apart from deleting the subrepo, commit and pulling it back, which seems very bad for the history and all-around unclean.

So If you have any suggestions, I'll be happy to hear from you.

Thanks, forry

admorgan:
Do you still have a copy of your workspace? If so zip it and attach it and I will be happy to investigate it. If you can't I understand it will just require more back and forth.

forry:
Yes, I have. I tested it and packed the inside of my repo. The subrepo is in the geGL subfolder.
[GPUE.zip](https://github.com/ingydotnet/git-subrepo/files/4000323/GPUE.zip)

admorgan:
[...]

Issue #350, opened 2018-03-07
fatal: Not a valid object name: ''.


Sharcoux:
Hi,

I am trying git subrepo for the first time, so maybe I just missed something in the doc.

I have a super project including 3 subrepos. I successfully created the subrepos. Now, I made some changes on my super project including some files from the subrepos. I want to push back thoses changes to the subrepos git projects.

I commited my changes at the super project level, and now, if I understood correctly, I should `git subrepo branch mySubRepo`, but I get:

```
git-subrepo: Command failed: 'git branch subrepo/client '.
fatal: Not a valid object name: ''.
```

I am using the release 0.4.0 on MacOS.

grimmySwe:
If there are no problems, you shouldn't need to run the `git subrepo branch mySubRepo` command. If nothing has happened in the subrepos `git subrepo push mySubRepo` should be enough. If something has happened it will complain and you need to run `git subrepo pull mySubRepo`first.

grimmySwe:
In further error reports, `-vd` is valuable flags that will give extra information in the debug process.

Sharcoux:
git subrepo pull client
Subrepo 'client' is up to date.

git subrepo push client
git-subrepo: Command failed: 'git branch subrepo/client '.
fatal: Not a valid object name: ''.

Sharcoux:
git subrepo push client -vd
>>> git rev-parse --verify HEAD
* Assert that working copy is clean: /.../
* Check for worktree with branch subrepo/client
  * Fetch the upstream: git@gitlab.com:... (master).
>>> git fetch --no-tags --quiet git@gitlab.com:... master
  * Get the upstream subrepo HEAD commit.
>>> git rev-parse FETCH_HEAD^0
  * Create ref 'refs/subrepo/client/fetch'.
>>> git update-ref refs/subrepo/client/fetch ...
* Check upstream head against .gitrepo commit.
* Deleting old 'subrepo/client' branch.
* Remove worktree:
>>> git branch -D subrepo/client
* Create subrepo branch 'subrepo/client'.
  * Check if the 'subrepo/client' branch already exists.
  * Subrepo parent: ...
  * Create new commits with parents into the subrepo fetch
>>> git rev-list --reverse --ancestry-path ...HEAD
  * Create branch 'subrepo/client' for this new commit set .
>>> git branch subrepo/client
git-subrepo: Command failed: 'git branch subrepo/client '.
fatal: Not a valid object name: ''.

[...]

PR #497, opened 2020-08-27
Avoid stale worktree on push

DanielBauten:
fix issue #496

PR #498, opened 2020-08-27
Fix error "fatal: Not a valid object name: ''."

DanielBauten:
# Problem

In case there are no (relevant) commits (e.g. directly after `git subrepo push`) which cause `subrepo:branch()` (i.e. on `git subrepo pull <subrepo>`) to fail with an git error message

> git-subrepo: Command failed: 'git branch subrepo/\<subrepo\> '.
> fatal: Not a valid object name: ''.

See issues #348, #350 and #424.

# Cause

This is because second parameter to `git branch` is an empty string as variable `prev_commit` has not been set. Thus, [branch creation](https://github.com/ingydotnet/git-subrepo/blob/a04d8c2e55c31931d66b5c92ef6d4fe4c59e3226/lib/git-subrepo#L789) fails in that case.

# Solution

To create the subrepo branch, nevertheless, it can be created on the cloned/pulled commit from `.gitrepo` file (per variable `subrepo_commit`) as there have not been any relevant changes.
Furthermore – to avoid unnecessary large range on the subsequent [call of `git filter-branch`](https://github.com/ingydotnet/git-subrepo/blob/a04d8c2e55c31931d66b5c92ef6d4fe4c59e3226/lib/git-subrepo#L800) – variable `first_gitrepo_commit` should be neutralized with `subrepo_commit` as well, so that variable `filter` results in an empty revision range.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -788,6 +788,7 @@ subrepo:branch() {
     if [[ -z "$prev_commit" ]]; then # i.e. there are no new reachable commits
       o "Create branch '$branch' for subrepo commit $subrepo_commit."
       prev_commit="$subrepo_commit"
+      first_gitrepo_commit="$subrepo_commit" # force neutralization of $filter to avoid unnecessary large range (see below)
     else
       o "Create branch '$branch' for this new commit set $prev_commit."
     fi
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -785,7 +785,12 @@ subrepo:branch() {
       fi
     done
 
-    o "Create branch '$branch' for this new commit set $prev_commit."
+    if [[ -z "$prev_commit" ]]; then # i.e. there are no new reachable commits
+      o "Create branch '$branch' for subrepo commit $subrepo_commit."
+      prev_commit="$subrepo_commit"
+    else
+      o "Create branch '$branch' for this new commit set $prev_commit."
+    fi
     RUN git branch "$branch" "$prev_commit"
   else
     o "No parent setting, use the subdir content."


admorgan:
Would you be willing to add some tests for this?

admorgan:
If this blindly "fixes" the issues you linked I think that could be a problem. Your example of a pull after a push is desirable, but when the ancestry has changed I feel we should acknowledge that there is a potential for your code not being what you expect. I have to dive deeper, but I believe it would not effect the included issues because we would fail before we reach this code due to the SHA not existing. @ingydotnet I would like you to look at this one if you have some time.

Overall thank you, this has been a thorn in my side. I have another favor to ask. We are currently working on the release/0.4.2 branch would you mind applying these patches there instead?

DanielBauten:
> Would you be willing to add some tests for this?

We were able to recreate the problem very well during the corrections to #497 and therefore believe that it was not just blindly fixed the issue linked. However, we have only been working with sub-repo for a few weeks and therefore have no experience in how to write suitable tests.

So – if @ingydotnet sees the need for this – it would be great if a more experienced developer could write some suitable tests.

blaggacao:
I think I hit this after a sqash merge of a feature branch where I did git subrepo on the feature branch. Does that sound plausible?

DanielBauten:
> I think I hit this after a sqash merge of a feature branch where I did git subrepo on the feature branch. Does that sound plausible?

Sounds plausible. If it reproducable in your case, perhaps you can check wether our fix is applicable on your situation.

natew:
Is there a way to fix this manually for now?

admorgan:
Modify the parent sha in your .gitrepo file to be the previous commit in your repo.

natew:
Unfortunately that seems to "squash" all the changes across a bunch of commits into one... We work on branches that rebase off master, then merge, but it seems every time i have to update the SHA to the latest, so all intermediate messages are being removed.

dwettstein:
Probably this PR refers to https://github.com/ingydotnet/git-subrepo/issues/503 and https://github.com/ingydotnet/git-subrepo/issues/532 as well.

jameswalmsley:
Can we merge this PR?

jcalfee:
What SHA do I use for the parent?  Is SHA easier to get from the client or the server?  Maybe I should git log on the subrepo folder to find this?

admorgan:
I have been working with this patch, and have found a number of paths that cause issues with regenerating the branch. I am addressing this through UI improvements and better error messaging instead of automatic regeneration for these reasons.

PR #500, opened 2020-09-08
Release/0.4.2

admorgan:
Improve testing

### DIFF
--- a/test/setup
+++ b/test/setup
@@ -8,6 +8,9 @@ set -e
 # Set the GIT_SUBREPO_ROOT for testing.
 source "$PWD"/.rc
 
+# Get the location of this script
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
 BASHLIB="$(
   find "$PWD"/ -type d -name bin -o -type d -name lib | tr '\n' ':'
 )"
@@ -18,8 +21,7 @@ source bash+ :std
 
 source 'lib/git-subrepo'
 
-# TODO: Move to test/tmp
-export TMP="tmp"
+export TMP="$SCRIPT_DIR/tmp"
 rm -fr $TMP
 
 #export GIT_EXEC_PATH="$PWD/lib:$(git --exec-path)"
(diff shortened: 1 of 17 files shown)


test/setup
```diff
@@ -6,20 +6,22 @@ git config core.autocrlf input
 set -e
 
 # Set the GIT_SUBREPO_ROOT for testing.
-source $PWD/.rc
+source "$PWD"/.rc
+
+# Get the location of this script
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
ingydotnet, line 12: Bash doesn't need quotes on assignments. These are the same:
```
x="$( ... )"
x=$( ... )
```
The `$( ... )` is parsed as a single expression even though it has ws in it. 

Please remove the quotes. 

At some point after the release, I'd like to go over all the Bash stuff, to bring it up to a set of standards that I've developed since starting subrepo.

test/setup
```diff
@@ -6,20 +6,22 @@ git config core.autocrlf input
...
+
+BASHLIB="$(
```
ingydotnet, line 14: Same with outer quotes. The inner quotes *are* needed.

test/setup
```diff
@@ -102,49 +104,49 @@ test-exists() {
...
           "Directory '$f' does not exist"
       else
-        ok "`[ ! -f "$f" ]`" \
+        ok "$([ ! -f "$f" ])" \
           "File '$f' does not exist"
       fi
     else
       if [[ $f =~ /$ ]]; then
-        ok "`[ -d "$f" ]`" \
+        ok "$([ -d "$f" ])" \
```
ingydotnet, line 115: This is better, but in bash `[[ ... ]]` is prefered to `[ ... ]`.
When you use `[[` you don't need to quote variables inside. 

There is one exception. The RHS of a comparison should be quoted if it has a variable expansion:
```
if [[ $a == "$b" ]]; then ...
```

ingydotnet (approved):
I made some comments. But overall I approve. This is the right direction.
There are still a lot of changes I'd like to make, but we can do it next release.

PR #505, opened 2020-11-11
My 2 cents on adding minimal support for LFS Release/0.4.2 lfs

sonatique:
This seems to work for me with subrepo repository having LFS enabled (Bitbucket).

This is only really minimalist, just for maybe helping someone or triggering a real dev around this.


polski-g:
Does not seem to work with Gitlab. I think its trying to pull down `roles/nrpe/files/nrpe.Debian.bin` instead of just `files/nrpe.Debian.bin` from the subrepo's repo.

```
git subrepo clone git@gitlab.local:/ansible/role-nrpe.git roles/nrpe
git-subrepo: Command failed: 'git read-tree --prefix=roles/nrpe -u ASDFASDF1234'.
Downloading roles/nrpe/files/nrpe.Debian.bin (216 KB)
Error downloading object: roles/nrpe/files/nrpe.Debian.bin (955871b): Smudge error: Error downloading roles/nrpe/files/nrpe.Debian.bin (GHJKGHJK5678): [GHJKGHJK5678] Object does not exist on the server or you don't have permissions to access it: [404] Object does not exist on the server or you don't have permissions to access it

Errors logged to ......./.git/lfs/logs/20220809T104241.815896285.log
Use `git lfs logs last` to view the log.
error: external filter 'git-lfs filter-process' failed
fatal: roles/nrpe/files/nrpe.Debian.bin: smudge filter lfs failed
```

Issue #525, opened 2021-02-20
test/encode.t fails when global .gitignore contains an entry matching a test path


falken42:
On macOS Big Sur 11.1 (20C69), the current `master` (commit `2f685964`) for git-subrepo fails a single test in test/encode.t (unrelated lines omitted for brevity):

```
Michaels-MBP:git-subrepo $ git --version
git version 2.30.0
Michaels-MBP:git-subrepo $ make test
prove  test/
test/branch-all.t ................ ok   
[...]
test/config.t .................... ok    
test/encode.t .................... 85/? #   Failed test 'subrepo push command output is correct'
#   at test/encode.t line 62.
#     got: 'Subrepo '~/......s:a^t?r a*n[g@{e.lock' has no new commits to push.'
#   expected: 'Subrepo '~/......s:a^t?r a*n[g@{e.lock' pushed to '/Users/falken/src/git-subrepo/test/tmp/upstream/bar' (master).'
test/encode.t .................... Failed 1/85 subtests 
test/error.t ..................... ok    
[...]
test/submodule.t ................. ok   

Test Summary Report
-------------------
test/encode.t                  (Wstat: 0 Tests: 85 Failed: 1)
  Failed test:  85
Files=36, Tests=337, 173 wallclock secs ( 0.15 usr  0.10 sys + 60.24 cusr 90.23 csys = 150.72 CPU)
Result: FAIL
make: *** [test] Error 1
```

Running the test suite on Ubuntu 20.04.2 LTS passes without any failures, so this looks to be specific to macOS.  zsh also appears to be unrelated, as the same test fails when running with bash.

admorgan:
What bash version are you using? I unfortunately don't have the facility to debug as I don't have any Apple products.

falken42:
Here's the details:

```
Michaels-MBP:git-subrepo $ uname -a
Darwin Michaels-MBP 20.2.0 Darwin Kernel Version 20.2.0: Wed Dec  2 20:39:59 PST 2020; root:xnu-7195.60.75~1/RELEASE_X86_64 x86_64
Michaels-MBP:git-subrepo $ bash --version
GNU bash, version 5.1.4(1)-release (x86_64-apple-darwin19.6.0)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
[...]

PR #527, opened 2021-02-21
[PR] Pass --force to git add so a user's global .gitignore does not affect tests

falken42:
This PR fixes an issue ( #525 ) where a git-subrepo test would fail if the test happened to use a path or filespec which matched an entry in a user's global .gitignore file.

By passing `--force` to `git add`, test files are added to the test repositories regardless of any .gitignore file settings, and allows the tests to pass as expected.

### DIFF
--- a/test/setup
+++ b/test/setup
@@ -68,7 +68,7 @@ add-new-files() {
   local file
   for file in "$@"; do
     echo "new file $file" > "$file"
-    git add "$file"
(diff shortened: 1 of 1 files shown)


admorgan:
Can you put this on the release/0.4.5 branch? If so I will accept it immediately.

falken42:
Done!

PR #536, opened 2021-06-15
Fix for #534

### LINKED ISSUE #534: git subrepo status doesn't show subrepos that have a common prefix
If you have 2 subrepos named "aaa" and "aa", running "git subrepo status" would only show the "aa" subrepo and omit "aaa".

cciocov:
Fix git subrepo status command for subrepos that share a common prefix. Closes #534.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -955,7 +955,7 @@ subrepo:status() {
     local _worktree
     _worktree=$(
       git worktree list |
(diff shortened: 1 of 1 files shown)


admorgan:
It would be great to have a test case added. If you move this to to release/0.4.5 I will cut a release with this patch.

admorgan:
Pushed onto release/0.4.5

PR #544, opened 2021-09-14
Improve error message to give possible explanations about why the command failed

ArnaudBienner:
I faced this error when trying to `git subrepo clone` a newly, empty repository, and it was not immediately obvious to me that the problem was that the repository was empty and thus has no branch.
I listed other possible reasons why the command would fail, not sure if there could be other ones.


admorgan:
If you put is on release/0.4.5 I will accept it.

ArnaudBienner:
Sorry I didn't have the time to look at this, and forgot about it.

I changed the base branch to release/0.4.5.

Let me know if this is OK for you now.

Issue #539, opened 2021-07-01
Parent not updated in .gitrepo file after certain merges


karptonite:
I've run into this a few times, and have now made a test repository to reproduce the issue:
https://github.com/karptonite/subrepotest1

Here is what happened. `subrepotest1` is a repository which contains `subrepotest2` as a subrepo.

Normally, when a change is made in `subrepotest2`, we then use `git subrepo pull` to pull the changes into `subrepotest1`. This pulls in the update, and updates both the `commit` and the `parent` in the `.gitrepo` file.

But there is a problem. Say two users are working in `suberepotest1`. One of them pulls in new changes from `subrepotest2` with git subrepo, and pushes them to the remote repository. The second user makes and commits some local changes in `subrepotest1`, then tries to push. There is a conflict, of course. The second user pulls from the remote with --no-rebase, and pushes again. All seems to be fine.

But then the next time a change is made on `subrepotest2`, and it is pulled into `subrepotest1` via git subrepo, git subrepo fails to update the `parent` in the `.gitrepo` file. In fact, as far as I can tell, git subrepo will never update `parent` again. This doesn't seem to prevent the subrepo pulls from continuing to work, but it DOES cause very bad perfomance, the more commits you are removed from the issue.

You can see this whole process, including the bug, (in this commit here: https://github.com/karptonite/subrepotest1/commit/d023d663182a0826872c20fc64daffcea77676fc) in the repo linked above, but if you need more information, I'm happy to provide it.

admorgan:
You are my best friend for the day. The problem is the merge combines the README.md and when git asks for history (using internal tricks because deep introspection would be prohibitively expensive), git says "Yep, there is a patch that has content from the folder you are interested in, and it has changes that didn't come from a squash merge". We then fall back to the really expensive merging function instead basically a FF via squash merge. This will take a bit of investigation on how to cheaply ensure we get the right answer.

I do appreciate the test I can add to our test cases after it is fixed.

karptonite:
FYI, I've "Fixed" this occasionally by making a new commit, updating the "parent" to the previous (well, current) commit. If nothing else, making `git subrepo pull -f` or `git subrepo clone -f`  rewrite the `.gitrepo` file even if it is otherwise up to date might help if the `.gitrepo` file gets into some unexpected state like this.

admorgan:
That is the current workaround. But I believe we can make this better.

eyeripaSICKAG:
[...]

PR #547, opened 2021-10-24
fix issue with .gitrepo parent field  not being updated when using pu…

tsadiko74:
fix issue with .gitrepo parent field  not being updated when using `git pull --rebase` (or if that option was set in git config) on a diverged parent repo

In case of a `git pull` using _rebase_ on the parent repository ,  the field subrepo.parent SHA-1 could be wrong.
That would cause an error on the next `git subrepo pull` command  , the workaround requires manual change of the .subrepo file, This proposed commit fixes that issue 

it may also be applicable for this reported issue -- https://github.com/ingydotnet/git-subrepo/issues/539#issue-934973979

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -1349,10 +1349,32 @@ read-gitrepo-file() {
 
   SAY=false OUT=true RUN git config --file="$gitrepo" subrepo.commit
   subrepo_commit=$output
+  
+  # In case of pull using rebase, the field subrepo.parent SHA-1 could be wrong.
+  #FAIL=false \
+  #SAY=false OUT=true RUN git config --file="$gitrepo" subrepo.parent
+  #subrepo_parent=$output
 
-  FAIL=false \
-  SAY=false OUT=true RUN git config --file="$gitrepo" subrepo.parent
-  subrepo_parent=$output
+  # Finding the right parent SHA-1
+  SAY=false OUT=true RUN git log --oneline --pretty=format:"%h"  -- "$gitrepo" 
+  gitrepo_commit_list=$output
+  found_comm="Empty"
+  for comm in $gitrepo_commit_list; do
+    parent_change_commit=$(
+      git diff -U0 $comm^! "$gitrepo" |
+        grep "parent = "
+    ) || true
+    if [ ! -z "$parent_change_commit" ]; then
+      found_comm=$comm
+      break
+    fi
+  done
+  if [ "$found_comm" != "Empty" ]; then
+    SAY=false OUT=true RUN git rev-parse "$comm^"
+    subrepo_parent=$output
+  else
+    subrepo_parent=
+  fi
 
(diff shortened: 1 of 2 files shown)


lib/git-subrepo
```diff
@@ -1349,10 +1349,32 @@ read-gitrepo-file() {
...
   subrepo_commit=$output
-
-  FAIL=false \
-  SAY=false OUT=true RUN git config --file="$gitrepo" subrepo.parent
-  subrepo_parent=$output
+  
+  # In case of pull using rebase, the field subrepo.parent SHA-1 could be wrong.
+  #FAIL=false \
+  #SAY=false OUT=true RUN git config --file="$gitrepo" subrepo.parent
+  #subrepo_parent=$output
```
admorgan, line 1356: Forgot to remove commented out old code

lib/git-subrepo
```diff
@@ -1349,10 +1349,32 @@ read-gitrepo-file() {
...
+
+  # Finding the right parent SHA-1
+  SAY=false OUT=true RUN git log --oneline --pretty=format:"%h"  -- "$gitrepo" 
+  gitrepo_commit_list=$output
+  found_comm="Empty"
```
admorgan, line 1361: This code base likes everything spelled out, I would prefer found_comm be renamed to common_sha or something fully spelled out.

lib/git-subrepo
```diff
@@ -1349,10 +1349,32 @@ read-gitrepo-file() {
...
+      git diff -U0 $comm^! "$gitrepo" |
+        grep "parent = "
+    ) || true
+    if [ ! -z "$parent_change_commit" ]; then
+      found_comm=$comm
+      break
+    fi
+  done
+  if [ "$found_comm" != "Empty" ]; then
+    SAY=false OUT=true RUN git rev-parse "$comm^"
```
admorgan, line 1373: I disagree that this would be the logical parent. I believe it would be the value + parent found in the  grep on line 1365.

admorgan (commented):
The big question is what state is the repository in after a rebase. I personally believe this will work out 99.9% of the time. I also don't know how to reliably determine the 0.1% it wouldn't. I feel like this is a positive change, but I am curious what @ingydotnet thinks.

trueqbit:
How does it go with this proposed PR? Recently I am constantly in the situation where I can't push changes anymore because git-subrepo thinks that there are no commits. I have to pull by force and reapply all the changes that the resulting commit reverted.

karptonite:
Just popping in to say that I'd love to see this get merged, if the maintainers think it solved the issue!

ingydotnet:
I'll talk to @admorgan about it soon.

Amatewasu:
(Just wanted to drop in and show my interest in this getting merged if the maintainers believe it's resolved the issue. Thank you a lot!)

PR #549, opened 2021-10-27
Push options support

queil:
This PR adds support for passing push options (as described in [the docs](https://git-scm.com/docs/git-push) when pushing subrepos changes.

Outstanding problems:
* [ ] Not sure how to enable passing multiple push options (not a bash expert here)
       Only the first option gets passed atm
* [ ] Not sure how to test it properly in the test suite
       At present I can only see the generated command if I unset `advertisePushOptions = true` on the bar repo
* [ ] Include the new option in auto-completion and man

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -61,26 +61,27 @@ See 'git help subrepo' for complete documentation and usage of each command.
 
 Options:
 --
-h           Show the command summary
-help        Help overview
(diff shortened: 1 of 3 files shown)


admorgan:
I am curious what options are you trying to pass to push?

queil:
For example https://docs.gitlab.com/ee/user/project/push_options.html

Issue #253, opened 2017-02-08
Add warning when combining arguments --branch and --all in subrepo pull


TinyTheBrontosaurus:
I have three subrepos (I'm calling them zords) I am tracking from a main repo (calling megazord). All the zords have the same branch names (`master` and `foo`). In my scenario, I want to change the tracking branches all the zords on megazord to `foo`. I tried to following command  

    git subrepo pull --branch=foo --all

which gave this output

    Subrepo 'the_zords/zord1' pulled from '../zord1' (master).
    Subrepo 'the_zords/zord2' pulled from '../zord2' (master).
    Subrepo 'the_zords/zord3' pulled from '../zord3' (master).

However, when I run these three commands separately, it does what I want.

    git subrepo pull --branch=foo the_zords/zord1
    Subrepo 'the_zords/zord1' pulled from '../zord1' (foo).

    git subrepo pull --branch=foo the_zords/zord2
    Subrepo 'the_zords/zord1' pulled from '../zord2' (foo).

    git subrepo pull --branch=foo the_zords/zord3
    Subrepo 'the_zords/zord1' pulled from '../zord3' (foo).

I assume the problem with the combining `--branch` and `--all` is that `--all` overrides it. That's fine behavior for my case; I don't mind doing all three commands as it's probably safer anyways. All I think is missing is that there should be a warning saying something along the lines of `Argument --branch is being ignored`

hever:
It is be illegal to use `--all` and `--branch` together? Or it is a bug and this should work? (Now it's not work - only with master branch).

admorgan:
I am going to declare that it is illegal to use --all and --branch together. The odds of you doing it on purpose are very low, and the utility is also low. I will add an error message and just return.

hever:
I think it could be legal. But maybe I'm looking to that in a bad way.

Our situation... We have project divided into subrepo. I want to work on some feature, so I will switch to new branch (on master project).

`git checkout -b new-feature`

My work will do the changes in some subrepos. Now I would like to push my changes to branches on subrepos.

`git subrepo push --all -b new feature` (imposible)

My colleague would like to see what I'm work on

```
git checkout -b new-feature
git subrepo push --all -b new feature  # imposible
```

Now I need to study on which subrepos I was doing changes and do the push for every subrepo. And my coleague needs to inspect every of suprepos if i do some changes here. Can please you tell me where is my workflow wrong?

admorgan:
[...]

PR #552, opened 2021-11-18
Add support --branch together with --all mentioned in #253

hever:


lib/git-subrepo
```diff
@@ -694,6 +690,16 @@ subrepo:fetch() {
     error "Can't fetch subrepo. Remote is 'none' in '$subdir/.gitrepo'."
   fi
 
+  if $all_wanted && [[ -n $subrepo_branch ]]; then
```
admorgan, line 693: Wouldn't something similar be necessary in subrepo:push?

admorgan (changes requested):
Independent of the technical issues I presented inline. I am concerned with a program overriding my explicit commands. I would be ok with a check to ensure that the branch existed on all effected repos and then executed accordingly or failed with a message telling me where it was missing, but to fall back to the previously configured branch after overriding it does not feel like an ok thing to do without direct user interaction. And as this run on multiple OS with multiple versions of bash getting user feedback is extraordinarily hard.

hever:
I'll try to understand you. I'm using subrepo in different use case for me it is meaningful to do this by this way.

Every command I'm using is somehow dangerous. This is life. This one could be dangerous also. But preventing users from doing stupid things is preventing them from doing the smart ones. This command potentially do what you might not want. But it is everything in git, everything is revertible. Need for user interaction is not the way IMHO. There must be good feedback and user must know what has been done.

lib/git-subrepo
hever: I'm not sure. I don't think anything is missing in subrepo:push. It's works fine with --all.

Issue #553, opened 2021-12-02
git subtree push to skip nested subrepos


michelemartone:
Hi.
I am aware of #417 and  #165 and #161 #82 and this ... is yet another query regarding 'nested subrepos' support :-)
Specifically, if both "bar" and "baz" in "bar/baz" are subrepos, then "git-subrepo push bar" pushes "baz" as well into the "bar" upstream.
Maybe it would make sense for "git-subrepo push" to skip directories containing ".gitrepo" (at least optionally)?

Background of this is to allow "git-subrepo" setups where one would lay subrepos according to the most natural order required by the packages, yet avoiding "excessive" pushes as described above.

michelemartone:
I've put together a patch (#554) for push to skip nested subrepos, with test.
I omitted updating documentation; probably having this as an option would be the best thing for a number of users.

PR #554, opened 2021-12-02
push to skip nested subrepos (as mentioned in Issue #553)

michelemartone:


jfrey-xx:
Hello,

I have been using git-subrepo for several personal and professional projects for a couple years. This is a *very* handy tool :+1:  I just came across a situation where I need the feature proposed in this PR (I want to keep several repositories separated and cannot avoid having one inside a subfolder of the other). Because I don't want to break my existing use of git-subrepo (where sometimes I *do* leverage on the fact that nested subrepo are pushed), I added an option to enable the "skip nested subrepos on push". This option is passed upon clone or init and resides within the `.gitrepo` file (it can also be updated with `git subrepo config`).

The option is (ill?)named "shallow" at the moment, (`-w` or `--shallow` flag upon clone or init).

The commit is probably rough on the edges, I just began using it (so far so good), the documentation and test suite should be updated, but it might interest others, and maybe help to get this PR merged at some point.

The code is here: https://github.com/jfrey-xx/git-subrepo/commit/55796941aa30c22f9af04410108b884c869f6a89

PR #560, opened 2022-02-24
Remove unintended space characters in Readme

cafhach:
Some mentions of arguments like `--message` were shown with an additional space character, e.g. `-- message` due to the linebreak within the code markup section


admorgan:
These files are automatically generated and the generator is the cause of the "broken" lines. The files reflowed correctly by the generator can be found in PR #563

PR #561, opened 2022-03-29
Fix subrepo path comparison

Greesb:
The path were compared like regex, the problem is that if we have 2 folders with the same starting name, the comparison would fail and one of the folder would be ignored (eg: `foobar` would be ignored if a previous `foo` subrepo existed)


admorgan:
Duplicate of #534 and addressed in #536

PR #580, opened 2023-04-14
Avoid stale worktree on push (2023 edition)

### LINKED ISSUE #496: stale worktree prevents push of new commits
# Problem

The command `git subrepo push <subrepo>` leads to an unexpected message

> Subrepo '\<subrepo\>' has no new commits to push.

if there are new commits present and there is also a stale, outdated worktree for the branch `subrepo/<subrepo>`.

See issues #326 and #352. The described condition can be caused e.g. as described in #347.

# Primary Cause

In case `git subrepo push <subrepo>` has indeed no new commits to push, the therefore created branch and worktree are not deleted by the subsequent [call of git:delete-branch](https://github.com/ingydotnet/git-subrepo/blob/a04d8c2e55c31931d66b5c92ef6d4fe4c59e3226/lib/git-subrepo#L666) since execution of `subrepo:push()` ends prematurely: https://github.com/ingydotnet/git-subrepo/blob/a04d8c2e55c31931d66b5c92ef6d4fe4c59e3226/lib/git-subrepo#L643

A subsequent call of `git subrepo push <subrepo>` then fails as described above.

# Secondary Cause

In `subrepo:push()`, a cleanup is already intended with a [call of git:delete-branch](https://github.com/ingydotnet/git-subrepo/blob/a04d8c2e55c31931d66b5c92ef6d4fe4c59e3226/lib/git-subrepo#L603). But since the global variable `worktree` is unset at this time, a stale worktree (caused by e.g. _Primary Cause_) is _not_ removed. Thus, the git command `git branch -D "subrepo/<subrepo>"` fails of course.

The subsequent call of `subrepo:branch` therefore returns without creating an updated history of the subrepo branch. As a result, the new commits are not included in this branch, so there are no known changes to the upstream repository.

# Solution

**Primary cause** would be fixed by a mandatory call of `git:delete-branch` before exiting `subrepo:push()` in every case. Note: This would also fix issue #347.

**Secondary cause** would be fixed by an early determination of the expected worktree path in `subrepo:push()`.

andreasxp:
Fixes #496

This is an updated pull request of the one @DanielBauten opened about 3 years ago (#497). I rebased it onto current master and provided a testcase.

@admorgan sorry for the ping, could you please take a look and merge if you're ok with the changes? I see that the last PR went dead for 2 years, and merging this would allow our company to adopt this tool instead of the awful submodules.

Thanks!

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -64,7 +64,7 @@ Options:
 h           Show the command summary
 help        Help overview
 version     Print the git-subrepo version number
- 
+
 a,all       Perform command on all current subrepos
 A,ALL       Perform command on all subrepos and subsubrepos
 b,branch=   Specify the upstream branch to push/pull/fetch
@@ -77,7 +77,7 @@ file=       Specify a commit message file
 r,remote=   Specify the upstream remote to push/pull/fetch
 s,squash    Squash commits on push
 u,update    Add the --branch and/or --remote overrides to .gitrepo
- 
+
 q,quiet     Show minimal output
 v,verbose   Show verbose output
 d,debug     Show the actual commands used
@@ -616,6 +616,8 @@ subrepo:push() {
     fi
 
     branch_name=subrepo/$subref
+    # We must make sure that a stale worktree is removed as well
+    worktree="$GIT_TMP/$branch_name"
     git:delete-branch "$branch_name"
 
     if $squash_wanted; then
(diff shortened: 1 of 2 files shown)


admorgan:
@andreasxp I am going to rebase this and push it. I am currently waiting for the tests to complete.

admorgan:
Rebased and fixed in 0.4.6

PR #614, opened 2024-03-25
Fixes issue #613 (support for GPG signed commits on upstream repo)

### LINKED ISSUE #613: No support for GPG signed commits on upstream repo
_"Remove the .gitrepo file"_ step filter includes commits that are not needed to be rewritten. Those commits are from upstream repo and if they are GPG signed rewriting drops the signature, changes hashes and later results in failing _push_.

edvinassabaliauskas:
Fixes #613

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -869,7 +869,7 @@ subrepo:branch() {
   local filter=$branch
   [[ $first_gitrepo_commit ]] && filter=$first_gitrepo_commit..$branch
   FAIL=false RUN git filter-branch -f --prune-empty --tree-filter \
-    "rm -f .gitrepo" "$filter"
+    "rm -f .gitrepo" -- "$filter" --first-parent
 
   git:create-worktree "$branch"
 


lib/git-subrepo
admorgan, line 1: Can you explain how this was effecting GPG signed commits? I am not able to see it.

edvinassabaliauskas: Yes, so the `filter-branch` is effectively rewriting all the commits even if there was no change made (no `.gitrepo` to remove). If filter encounters a merge commit (with more than one parent), continuing the filter on a second parent rewrites those commits. If commits are GPG signed during rewrite the signatures gets dropped which results in changed commit hash. This results in inability to push the branch to the remote, as local and remote histories do not match.
We actually want to remove `.gitrepo` files only from the newly created commits and not from the ones that are already on the remote.

I hope I've explained it clear enough 😃 It took me a while to understand what was happening with the commit rewriting.

PR #615, opened 2024-03-26
Adds the ability to --gpg-sign new commits to the upstream

### LINKED ISSUE #613: No support for GPG signed commits on upstream repo

edvinassabaliauskas:
- Added new option `--gpg-sign` to `branch`, `pull`, `push` commands.
- Also fixes #613 as it is related for the feature to work correctly.

### DIFF
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -77,6 +77,7 @@ file=       Specify a commit message file
 r,remote=   Specify the upstream remote to push/pull/fetch
 s,squash    Squash commits on push
 u,update    Add the --branch and/or --remote overrides to .gitrepo
+gpg-sign?   GPG-sign commits. The argument is optional and defaults to the committer identity; if specified, it must be stuck to the option without a space.
  
 q,quiet     Show minimal output
 v,verbose   Show verbose output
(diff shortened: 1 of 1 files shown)


kuwannet:
I’ve been playing with this and have some comments…

1. It would be good to honor whether `commit.gpgsign=true` has been set via `git config` and automatically pass `-S` or `--gpg-sign` to the commands that need it. Per the [git documentation](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits):

    > To configure your Git client to sign commits by default for a local repository, in Git versions 2.0.0 and above, run `git config commit.gpgsign true`. To sign all commits by default in any local repository on your computer, run `git config --global commit.gpgsign true`.
 
    When you have `commit.gpgsign=true` set and are used to having commits always signed automatically, it’s easy to forget to add the option to the `git subrepo …` commands. 
 
2. It would also be nice to use the `-S` option as a synonym for `--gpg-sign` as is supported by git itself:
 
    ```
    -S[<keyid>], --gpg-sign[=<keyid>], …
        GPG-sign commits. The keyid argument is optional and defaults to the committer identity; if specified, it must be stuck to the option
        without a space…
    ```

PR #620, opened 2024-05-30
feat: Add --commit option to git subrepo pull command

giladsh1:


admorgan:
I don't see how this is better than using `git subrepo pull --branch <sha>` which already works.

lib/git-subrepo.d/help-functions.bash
admorgan, line 1: help-function.bash is a generated file. These changes need to be made at the source of that generation.

lib/git-subrepo
admorgan, line 1: `git subrepo pull --commit` would override any `--branch` option. There isn't any comments to the user or the application indicting this is by design.

admorgan (changes requested):
I would also like to know why you feel this is better than the solution that already works of using `--branch <SHA>` to do the same thing. If you would like to keep this option it should also mention somewhere that `git subrepo pull` is effected by this option.

Issue #271, opened 2017-05-29
Install on windows instructions


carlokok:
On Windows one does generally use git bash but just git as a global command in path. I see there's a method to install subrepo as a sub command of git via Makefile but Make is generally not installed on Windows. A set of instructions on how to install it manually on Windows would be really useful.

skelliam:
https://github.com/ingydotnet/git-subrepo/issues/209

skelliam:
I can comment on this.  Mostly this is in the installation instructions.

1. Clone git-subrepo.  Checkout the release branch (currently release/0.4.0)
2. In control panel --> system --> advanced --> environment variables, create a new environment variable GIT_SUBREPO_ROOT and set it to the location where you cloned git-subrepo
3. While you are adding environment variables, also update your PATH to include the location of git-subrepo *and* the location of git-subrepo/lib
4. Close any open shell windows and reopen them, or to be really sure that you get the updated path, reboot.
5. Open a fresh shell window (cmd.exe) and try "git subrepo".

ConradBraam:
do we need/have a PR to ammend the docs as per @skelliam suggestion?

erikMagnussonSoftware:
@skelliam's instructions work and should make it to the main README.

bryanlarsen:
powershell command line:

```
$Env:GIT_SUBREPO_ROOT = pwd
$Env:PATH += ";${pwd}\lib"
```

PR #623, opened 2024-06-23
Initial patchset for integration with debian

spog:
In the process of preparing a debian git-subrepo package, a set of changes had to be prepared to adapt git-subrepo to debian runtime and build system requirements. 
Changes have been grouped as follows:
  * Removed hashbangs from sourced-only files (debian policy)
  * Removed executable permission on sourced-only files (debian policy)
  * Fixed bash-completion integration with git (git sub[tab][tab] did not find subrepo)
  * Drop unnecessary subdir in usr/share/git-subrepo (only for 'make install' installation method)
  * Fixed 'make test' for debian build (debian build of package runs from a non-git git-subrepo project and for different languages)
  * Generate test git-repos instead of committed binaries (i prefer text files in a git project whenever possible)

### DIFF
--- a/Intro.pod
+++ b/Intro.pod
@@ -78,7 +78,7 @@ Submodules tend to receive a lot of bad press. Here's some of it:
 
 =item * L<http://ayende.com/blog/4746/the-problem-with-git-submodules>
 
-=item * L<http://somethingsinistral.net/blog/git-submodules-are-probably-not-the-answer/>
+=item * L<https://web.archive.org/web/20171101202911/http://somethingsinistral.net/blog/git-submodules-are-probably-not-the-answer/>
 
 =item * L<http://codingkilledthecat.wordpress.com/2012/04/28/why-your-company-shouldnt-use-git-submodules/>
 
--- a/doc/intro-to-subrepo.swim
+++ b/doc/intro-to-subrepo.swim
@@ -61,7 +61,7 @@ article is about.
 Submodules tend to receive a lot of bad press. Here's some of it:
 
 * http://ayende.com/blog/4746/the-problem-with-git-submodules
-* http://somethingsinistral.net/blog/git-submodules-are-probably-not-the-answer/
+* https://web.archive.org/web/20171101202911/http://somethingsinistral.net/blog/git-submodules-are-probably-not-the-answer/
 * http://codingkilledthecat.wordpress.com/2012/04/28/why-your-company-shouldnt-use-git-submodules/
 
 A quick recap of some of the good and bad things about submodules:


admorgan:
I am in the middle of tracking down a couple of annoying bugs. The intentions of this patch look reasonable, but it may be a bit before I dig in and determine if I can approve it. This is largely due to the fact the bugs I am tracking down are caught by our CI and therefore it makes it harder to test these changes.

spog:
Thanks for the reply and take your time:)

admorgan (changes requested):
These are all auto-generated files, see the comment "# DO NOT EDIT. This file generated by pkg/bin/generate-help-functions.pl." present in each. The source for these files needs to be updated and then run `make compgen` to regenerate the files.

admorgan (changes requested):
I am sorry, but I completely stepped on this patch with the 0.4.7 release. If there is a list somewhere about what rules are expected for Debian tests I could help rework this patch.

admorgan (changes requested):
I would really like to have a test for both of these methods of install, but I am not sure how to test auto-completion. Any ideas?

lib/git-subrepo.d/help-functions.bash
admorgan, line 1: This is an autogenerated file, see the comment "# DO NOT EDIT. This file generated by pkg/bin/generate-help-functions.pl." present in each. The source for these files needs to be updated and then run make compgen to regenerate the files.

pkg/bin/generate-completion.pl
admorgan, line 1: This is an autogenerated file, see the comment "# DO NOT EDIT. This file generated by pkg/bin/generate-help-functions.pl." present in each. The source for these files needs to be updated and then run make compgen to regenerate the files.

pkg/bin/generate-help-functions.pl
admorgan, line 1: This is an autogenerated file, see the comment "# DO NOT EDIT. This file generated by pkg/bin/generate-help-functions.pl." present in each. The source for these files needs to be updated and then run make compgen to regenerate the files.

share/completion.bash
admorgan, line 1: This is an autogenerated file, see the comment "# DO NOT EDIT. This file generated by pkg/bin/generate-help-functions.pl." present in each. The source for these files needs to be updated and then run make compgen to regenerate the files.

test/status.t
admorgan, line 1: I am sorry, but the 0.4.7 release changed all this (hopefully to a state you don't need this version any longer). I appologize.

lib/git-subrepo.d/help-functions.bash
admorgan: Forgot that commenting on a commit doesn't give context. Moved the comment to the file instead of the commit.

lib/git-subrepo
```diff
@@ -12,21 +12,8 @@ set -e
...
 # Import Bash+ helper functions:
-SOURCE=${BASH_SOURCE[0]}
-while [[ -h $SOURCE ]]; do
-  DIR=$( cd -P "$( dirname "$SOURCE" )" && pwd )
-  SOURCE=$(readlink "$SOURCE")
-  [[ $SOURCE != /* ]] && SOURCE=$DIR/$SOURCE
-done
-SOURCE_DIR=$(dirname "$SOURCE")
-
-if [[ -z $GIT_SUBREPO_ROOT ]]; then
```
admorgan, line 23: GIT_SUBREPO_ROOT is advertised as a valid install method for Windows. See https://github.com/ingydotnet/git-subrepo/issues/271#issuecomment-373722235. Need to investigate if this will work with a Windows system.

spog:
@admorgan Hi, and thanks for the review. i am currently on vacation until the end of July. I hope we can proceed then.
regards, Samo

Makefile
admorgan, line 1: I agree that the test "source" repos should be generated by the tests instead of having their binaries in the repo. This solution doesn't work for me because the tests are frequently run outside the scope of the Makefile. Therefore this specific solution doesn't work. See the documentation and search for "prove". This logic needs to be moved to the test/setup script.

test/status.t
spog: No need to apologize and thanks for the info.

lib/git-subrepo
spog: I imagine, compatibility with Windows is a challenge. I can not say more at this point.

Makefile
spog: I'll try to address this according to your proposal.

spog:
@admorgan 
> I would really like to have a test for both of these methods of install, but I am not sure how to test auto-completion. Any ideas?

I've been trying to figure out how to test auto-completion specifically for git and all i've come up is the script below, which you may call with params otherwise added to the 'git' command and the script outputs completion words in a similar manner as a real interactive bash auto-completion. The script (git-completion-test.sh) is far from perfect, but i hope it may help:
`#!/bin/bash

source /usr/share/bash-completion/bash_completion
__load_completion git

function echo_vars ()
{
	local idx=0
	echo "cur=${cur}"
	while [ -n "${words[idx]}" ]; do
		echo "words[${idx}]=${words[idx]}"
		((idx++))
	done
	echo "cword=${cword}"
	echo "prev=${prev}"
}

COMPREPLY=()
words=()
cur=
words[0]=git
cword=1
prev=git

for arg_word in $(echo "${@}"); do
	if [ -z "${arg_word}" ]; then
		break;
	fi
	words[cword]=${arg_word}
	prev=${words[((cword-1))]}
	((cword++))
done
cur=${words[(($cword-1))]}
words[(($cword-1))]=
((cword--))
prev=${words[(($cword-1))]}
#echo_vars
#echo "Check as incomplete last word: ${cur}"
__git_main

if [ -z "${COMPREPLY[0]}" ] || [ "${COMPREPLY[0]}" == "${cur} " ]; then
	words[cword]=$cur
	prev=$cur
	cur=
	((cword++))
	#echo_vars
	#echo "Re-check as complete last word: ${prev}"
	__git_main
fi

if [ -z "${COMPREPLY[0]}" ] || [ "${COMPREPLY[0]}" == "${prev} " ]; then
	#echo "No words available to complete!"
	exit 1
fi

#echo "Words available to complete:"
idx=0
while [ -n "${COMPREPLY[idx]}" ]; do
	echo -n "${COMPREPLY[idx]}"
	((idx++))
done
echo
`
best regards, Samo

Makefile
spog: I prepared a new change by moving my Makefile changes from last two commits of this PR (regarding generation of test repos and potentially initializing the whole project as git project for cases), into the test/setup script.
Currently i have the new change only locally as modified last two commits of what constitutes this PR. But now i wander, if i force push the change into my git-subrepo fork, the original PR reference is lost. So how would you prefer to see this change presented with regard of the state of this PR?

spog:
@admorgan, sorry for requesting a review. I am not always sure what i am clicking in github:(

admorgan:
@spog, no worries, as is evident from my code review I have a lot of issues with their interface also. In case you have to work in github more frequently, if you do a force push to the same branch you have requested a PR on, it will automatically update the PR with the new code and even show the project what changed between the pushes.

The script is really useful and I thank you for helping with that. I am trying to expand it to ensure that "git subrepo ini" completes to the appropriate expansion and I will be adding a test for that as soon as I have some time to sort it out.

In short, Github kind of sucks as a UI; you can submit updates to this patch by force pushing on the same branch in your project and Github does something relatively sane; and I appreciate all the work you are putting into this.

PR #630, opened 2024-08-18
Add information about usage with nested subrepos

### LINKED ISSUE #629: Document intended usage with nested subrepos
I'm testing out git-subrepo for our project, which has currently 2 levels of subrepos, like this:
```
app
|- foo
    |- bar
```

This works, of course, as it's even simpler than the structure in git-subrepo itself. 
However what's not fully clear is the intended workflow. An example:

I open the `app` folder. I make a change in `bar`. I commit, and now I want to propagate this change.

what should I do?
### option 1:  `git subrepo push foo/bar`

This fails with:
```
Command failed: 'git branch subrepo/foo/bar '.
fatal: not a valid object name: '
```

### option 2: `git subrepo push foo` and then `git subrepo push foo/bar`

This fails for the same reason as above

### option 3: the only working way I've found

```
git subrepo push foo
cd ~/repos/foo #this is where the "real" clone is stored
git pull
git subrepo push bar
cd - #back to the top repo
git subrepo pull foo #necessary because the previous push added a commit
```

Is this the "intended" way? Looking at the `.gitrepo` files, I think I understand why this is necessary, but I feel like I'm maybe missing something

If not, it might be worth adding this to the README. I can make a pr for that

### option 4: just don't

It may be that the intended workflow is to make the changes in the "native" repos, and use way more `pulls` than `pushes`? This would also be fine, I just got excited at the possibility to work across projects as if they were a monolithic bundle.

abeforgit:
While testing git-subrepo, I found some gotchas with regards to nested subrepos. With a bit of thinking, the way git-subrepo acts is logical, but maybe not immediately intuitive.

This PR adds a section on working with nested subrepos to the readme. I hope it explains the logic in a clear manner.

Please let me know if I made some mistakes, I've only really just begun to try out git-subrepo, the chances I got something wrong are high.

[rendered](https://github.com/abeforgit/git-subrepo/blob/docs/add-info-about-nested-subrepos/ReadMe.pod#working-with-nested-subrepos)

closes #629


doc/git-subrepo.swim
```diff
@@ -599,6 +599,74 @@ If you want to chat about the `git-subrepo` command, join `#gitcommands` on
...
+    |- test-more-bash
+      |- bashplus
+      |- test-tap-bash
+
+However, it's important to understand how this works. Here's the key idea:
+*`git-subrepo` is 100% agnostic about your dependencies!*
+
+There is *no* special handling for nested subrepos. When you clone a
+subrepo, *all* `git-subrepo` does is download the code and set up the subrepo
+file. It couldn't care less about the actual contents of your code.
```
admorgan, line 625: "It couldn't care less about the actual contents of your code." Feels both a little harsh and a lot personified. Could this be removed and the following sentence reworked that I feel describes what is happening quite well?

doc/git-subrepo.swim
```diff
@@ -599,6 +599,74 @@ If you want to chat about the `git-subrepo` command, join `#gitcommands` on
...
+The structure is therefore as follows:
+
+  git-subrepo
+    |- bashplus
+    |- test-more-bash
+      |- bashplus
+      |- test-tap-bash
+
+However, it's important to understand how this works. Here's the key idea:
+*`git-subrepo` is 100% agnostic about your dependencies!*
```
admorgan, line 621: I feel for some, especially those that aren't English speakers as a primary language may have issues understanding what 100% agnostic means and what is being referred to by dependency. I feel that the subsequent paragraph describes what is happening enough that the paragraph starting on line 620 is superfluous, and could be safely removed.

doc/git-subrepo.swim
```diff
@@ -599,6 +599,74 @@ If you want to chat about the `git-subrepo` command, join `#gitcommands` on
...
+Let's look at what this means in practice. Say you have a structure as follows:
+
+  app-foo
+    |- barlib
+       |- bazlib
+
+You're working on app-foo, and make some changes to bazlib. How should you
+upstream these changes?
+
+Well, first you must realize that from the perspective of app-foo, you've simply
```
admorgan, line 638: I try to avoid using phrases such as "first you must realize". Even if it is true the fact that they are reading the documentation means that the documentation will educate them to the level they will realized these facts when they are done.

doc/git-subrepo.swim
```diff
@@ -599,6 +599,74 @@ If you want to chat about the `git-subrepo` command, join `#gitcommands` on
...
+made some changes to barlib. Whether or not those changes were themselves in a
+subrepo is irrelevant. So, just like any other changes, you run
+
+`git subrepo push barlib`
+
+Now, if `barlib` is a library you don't maintain, your responsibility would end
```
admorgan, line 644: I feel this is very well written and expresses how subrepo views ownership very well.

doc/git-subrepo.swim
```diff
@@ -599,6 +599,74 @@ If you want to chat about the `git-subrepo` command, join `#gitcommands` on
...
+here. `barlib`'s maintainer would see your changes and decide what to do with
+them. 
+
+If you do maintain `barlib`, you might now want to upstream the changes all the
+way into `bazlib`. You could be tempted to try running something like:
+
+`git subrepo push barlib/bazlib`
+
+but you'll soon find out that doesn't work. Why not? Simple: `app-foo` doesn't
```
admorgan, line 653: I would prefer to remove the word Simple. If all of this was simple it wouldn't take so many lines of explanation :)

doc/git-subrepo.swim
```diff
@@ -599,6 +599,74 @@ If you want to chat about the `git-subrepo` command, join `#gitcommands` on
...
+know anything about the link between `barlib` and `bazlib`.
+
+What you should do is treat the changes to `barlib` as if they would come from
+another contributor, aka:
+
+* go to your local copy of `barlib`
+* pull down the changes with a simple `git pull`
```
admorgan, line 660: I would like to remove the word simple. For all we know there will be merge conflicts and the like.

doc/git-subrepo.swim
```diff
@@ -599,6 +599,74 @@ If you want to chat about the `git-subrepo` command, join `#gitcommands` on
...
+
+`git subrepo push barlib/bazlib`
+
+but you'll soon find out that doesn't work. Why not? Simple: `app-foo` doesn't
+know anything about the link between `barlib` and `bazlib`.
+
+What you should do is treat the changes to `barlib` as if they would come from
+another contributor, aka:
+
+* go to your local copy of `barlib`
```
admorgan, line 659: I would like to propose changing "your local copy" to "a local copy", it is a minor difference, but I read your local copy as implying that you already have one, and "a local copy" as you should acquire one if you don't already have one.

doc/git-subrepo.swim
```diff
@@ -599,6 +599,74 @@ If you want to chat about the `git-subrepo` command, join `#gitcommands` on
...
+* pull down the changes with a simple `git pull`
+* you notice the changes include some work on your subrepo, time to upstream
+  them:
+  `git subrepo push bazlib`
+
+And you're done! One final step you'll likely want to do is to go back to
```
admorgan, line 665: I have been wondering about the "final step", while you would get the commit that was created by the push, none of your code would change and therefore there isn't a real benefit to doing the update. In face I have a todo to remove the commit from a push in the future to help with things like working in Gerrit or other code review tools that break some of the push assumptions that I have to use daily.

admorgan (changes requested):
I want to complement you on a very nice job, notably improving the documentation for this project. I really appreciate it. I do have some nit picks, but I couldn't have done this well myself.

abeforgit:
Oh this dropped off my radar, busy week
But thank you for the feedback, some good points, I'll update asap

abeforgit:
I think I (finally :see_no_evil:) addressed all your remarks, let me know if you want me to change something about the "final step"

doc/git-subrepo.swim
admorgan: I don't see being addressed in the commits on this PR.

admorgan: It has been a while so I am going to take the liberty of making this change myself so I can get the rest of your changes in.

PR #640, opened 2024-11-16
Fixed git-subrepo and tests if set ff=only

spog:
Some tests failed, when ff_only is set in users git configuration; like:
...
test/branch-rev-list-one-path.t .... Died at line 23 in main of test/branch-rev-list-one-path.t
test/branch-rev-list-one-path.t .... No subtests run
test/branch-rev-list.t ............. Died at line 26 in main of test/branch-rev-list.t
test/branch-rev-list.t ............. No subtests run ...

and git configuration is set like:
    $ git config --global pull.ff only
    $ git config --global merge.ff only

There may be a similar cause of failure described in issue 550.


Makefile
```diff
@@ -46,6 +46,7 @@ help:
 
 .PHONY: test
 test:
+	@echo uname: '$(shell uname)'
```
admorgan, line 49: Why did you feel this would beneficial?

test/00-git-config.t
```diff
@@ -0,0 +1,55 @@
...
+# Get git-subrepo project top directory
+PROJ_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )
+
+if [ -z "${PROJ_DIR}" ] || [ "${HOME}" != "${PROJ_DIR}" ]; then
+  is "${HOME}" "${PROJ_DIR}" \
+  "To define project-wide GIT setup for all tests: HOME '${HOME}' should equal PROJ_DIR '${PROJ_DIR}'"
+else
+
+  # Real GIT configuration for tests is set here:
+  rm -f "${PROJ_DIR}/.gitconfig"
```
admorgan, line 20: I am not ok with removing the .gitconfig from a person's working directory.

admorgan (changes requested):
I don't feel this is the best option for testing this feature. I believe that it would be sufficient to have a specific test that incorporates the fast forward only instead of setting it for all tests.

Makefile
spog: Hi and thanks for the question. Namely, while i was wrestling with github pipelines for Linux and Macos, i found this additional pipeline output very useful, especially in PR #639. So i thought i should leave it in this PR as well. There is no other value to it otherwise.

test/00-git-config.t
spog: Hi again,
i thought i made sure, that HOME is temporarily set to $PWD in 'test/setup' for the duration of running tests and only when git-subrepo's clone directory (PROJ_DIR) and HOME match, then rewrite tests-defined git configuration in PROJ_DIR.
regards, Samo

spog:
> I don't feel this is the best option for testing this feature. I believe that it would be sufficient to have a specific test that incorporates the fast forward only instead of setting it for all tests.

That may certainly be true, however i wanted to provide a way to define/set default tests-specific git configuration and is independent of current user's git config,

PR #650, opened 2025-05-10
Updated GPLv2 notice mostly to fix old fsf address

spog:
Please consider changing the GPLv2 notice in share/git-completion.bash to comply with its current wording (changed FSF address).

best regards, Samo


admorgan (changes requested):
I agree the license needs to be updated. This is a copy of the file from https://github.com/git/git/blame/master/contrib/completion/git-completion.bash. The entire file should be updated with the fixes applied instead of just the license updated.

PR #659, opened 2025-09-17
Added support to specify '-b branch@commit'

spog:
Current implementation already allows for specifying explicit commits like '-b commit' to pin subrepo sources (i.e during subrepo clone/pull). In that case, it also sets the name of the branch in the subrepo merge commit message and inside .gitrepo file to that commit hash. This addition allows for specifying both branch and commit. During the subrepo fetch phase, the commit is checked against the specified branch. If the requested commit can be reached from the head of the requested branch, the subrepo is re-fetched to that commit.
Afterwards command processing remains the same, except that given branch name is written to the subrepo clone/pull commit message and into the .gitrepo file.
The commit part of the '-b branch@commit' request may be abbreviated to the reasonable extent.

### DIFF
--- a/test/use-branch-at-commit.t
+++ b/test/use-branch-at-commit.t
@@ -18,7 +18,7 @@ clone-foo-and-bar
     "$OWNER/bar/Bar"
 }
 
-# Make a new branch bar1 in bar 
+# Make a new branch bar1 in bar
 (
   cd $OWNER/bar
   git checkout -b bar1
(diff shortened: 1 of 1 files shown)


admorgan:
Would you be willing to add tests for this?

spog:
Thanks for the suggestion.

spog:
Hi, i prepared initial tests for using branch@commit. I tried to follow your test scenario suggestions:)

admorgan:
I haven't looked really close because they are failing. Look at the output to see suggestions from shellcheck on what I believe will address most of the issues.

spog:
@admorgan hi, thanks for the reply. I fixed the pre-commit check, however tests are still failing (encode.t). Not on my behalf, i believe:)

Issue #323, opened 2017-12-01
git subrepo commit leaves a mess if it can't complete successfully


chrisdtaylor:
If anything goes awry prior to the file git-commit performed by subrepo commit, a mess is left behind.  I would suggest adding a "git reset --hard" or something similar to have that command clean up after itself if things don't work out.

grimmySwe:
@chrisdtaylor What version of subrepo are you using?
in version 0.4.0 we use worktree:s for the operations, so if anything goes seriously wrong you can always `git subrepo clean <subrepo>`

0.4.0 is currently on release/0.4.0 branch. We are working on it to get it to master.

chrisdtaylor:
Working with the release/0.4.0 branch.  I'm just referring to the case where the final git subrepo commit step fails.  Since the lead up to the actual commit in the multirepo involves removing the subrepo subdir and replacing it with the contents of the tmp worktree, if anything goes wrong in those steps, your multirepo worktree is potentially dirty.  Maybe it's just my work environment, but we have some pre-commit hooks that perform checks that can fail, leaving the multirepo default worktree dirty and in need of a git reset --hard.

grimmySwe:
@chrisdtaylor Good comments, there are some vulnerable points, so there is potential for some help if things go bad at the wrong time.

PR #660, opened 2025-10-26
Default to subrepo option to use --no-verify

### LINKED ISSUE #193: Make bypassing the pre-commit and commit-msg hooks optional
I'm using pre-commit hooks in my git project that validate code style and syntax (for puppet manifests). I noticed that this can break a `git subrepo clone` if the cloned repo itself contains i.e. unlinted stuff (and makes cloneing incredible slow, this example took 30s until failing).

```
 --- git/leap_platform ‹subrepo› » time git subrepo clone -v git@gitlab.com:shared-puppet-modules-group/mod_security puppet/modules/mod_security
* Determine the upstream head branch.
* Fetch the upstream: git@gitlab.com:shared-puppet-modules-group/mod_security (master).
  * Fetch 'git@gitlab.com:shared-puppet-modules-group/mod_security' (master).
  * Get the upstream subrepo HEAD commit.
  * Create subrepo remote 'subrepo/puppet/modules/mod_security' -> 'git@gitlab.com:shared-puppet-modules-group/mod_security'.
  * Create ref 'refs/subrepo/puppet/modules/mod_security/fetch'.
* Make the directory 'puppet/modules/mod_security/' for the clone.
* Commit the new 'puppet/modules/mod_security/' content.
  * Check that '222d510075b3869e2b78ea70b23780602267317b' exists.
  * Make sure '222d510075b3869e2b78ea70b23780602267317b' contains the upstream HEAD.
  * Put remote subrepo content into 'puppet/modules/mod_security/'.
  * Put info into 'puppet/modules/mod_security/.gitrepo' file.
  * Commit to the 'subrepo' branch.

git-subrepo: Command failed: 'git commit -m git subrepo clone git@gitlab.com:shared-puppet-modules-group/mod_security puppet/modules/mod_security

subrepo:
  subdir:   "puppet/modules/mod_security"
  merged:   "222d510"
upstream:
  origin:   "git@gitlab.com:shared-puppet-modules-group/mod_security"
  branch:   "master"
  commit:   "222d510"
git-subrepo:
  version:  "0.3.0"
  origin:   "https://github.com/ingydotnet/git-subrepo.git"
  commit:   "cb2995b"'.
git subrepo clone -v git@gitlab.com:shared-puppet-modules-group/mod_security   6.26s user 0.54s system 22% cpu 30.466 total
   
```

issueing `git commit -m "git subrepo clone git@gitlab.com:shared-puppet-modules-group/mod_security puppet/modules/mod_security"` showed me that it ran the pre-commit hook on every file and failed because of unlinted stuff.

Using `git commit --no-verify` worked for me with this patch:

```
--- git/git-subrepo ‹master* M› » git diff
diff --git a/lib/git-subrepo b/lib/git-subrepo
index 3c6f575..a121320 100755
--- a/lib/git-subrepo
+++ b/lib/git-subrepo
@@ -778,7 +778,7 @@ subrepo:commit() {
 
   o "Commit to the '$original_head_branch' branch."
   if [[ $original_head_commit != none ]]; then
-    RUN git commit -m "$(get-commit-message)"
+    RUN git commit --no-verify -m "$(get-commit-message)"
   else
     # We had cloned into an empty repo, side effect of prior git reset --mixed
     # command is that subrepo's history is now part of the index. Commit

```

So I'm wondering what to do here, I see three options:

- always use `git commit --no-verify`
- an optional parameter to `git subrepo clone` like `--no-verify`
- or an even more generic parameter like `--git-commit-options` that is passed directly to `git commit`

soraxas:
This closes #193, and potentially adresses #323

Since the clone / pull / ... command in `git subrepo` are operating on a different repo, it often doesn't make much sense to trigger the hook, e.g., pre-commit hooks in the main repo.

This PR defaults to use `--no-verify` flag on these related operation, and the behaviour can also be configured via

The `.gitrepo` file:
```ini
[subrepo]
    ...
    verify = true  # Enable hooks for this subrepo
```


lib/git-subrepo
```diff
@@ -531,7 +539,10 @@ subrepo:init() {
 
   o "Commit new subrepo to the '$original_head_branch' branch."
   subrepo_commit_ref=$original_head_commit
-  RUN git commit -m "$(get-commit-message)"
+  local no_verify_flag=' --no-verify'
+  $verify_wanted && no_verify_flag=''
+  # shellcheck disable=2086
+  RUN git commit$no_verify_flag -m "$(get-commit-message)"
```
admorgan: To prevent disabling 2086 this should use Parameter Expansion https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html#Shell-Parameter-Expansion-1

lib/git-subrepo
```diff
@@ -919,12 +934,17 @@ subrepo:commit() {
...
   o "Commit to the '$original_head_branch' branch."
   if [[ $original_head_commit != none ]]; then
     if [[ $commit_msg_file ]]; then
-      RUN git commit $edit_flag --file "$commit_msg_file"
+      # shellcheck disable=2086
+      RUN git commit$no_verify_flag $edit_flag --file "$commit_msg_file"
     else
-      RUN git commit $edit_flag -m "$commit_message"
+      # shellcheck disable=2086
+      RUN git commit$no_verify_flag $edit_flag -m "$commit_message"
```
admorgan: To prevent disabling 2086 this should use Parameter Expansion https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html#Shell-Parameter-Expansion-1

soraxas:
The issue was not on the pre-commit hook inside the subrepo, but the one in the main repo.

E.g., say there's a per-commit hook in the main repo, and say perhaps there's a python code formatter there

```
.
├── app
│   ├── .pre-commit-config.yaml
│   ├── subrepo1
│   ├── docs
│   ├── subrepo2
...
```
If I were to now clone a subrepo
```sh
git subrepo clone SOME_REPO_WITH_PYTHON_CODE new_subrepo
```

Then if the python code formatter did changed the newly cloned subrepo, the whole repo is now at a dirty state.
e.g.
```sh
git-subrepo: Command failed: 'git commit -m git subrepo clone ../py

subrepo:
  subdir:   "py"
  merged:   "12641cf"
upstream:
  origin:   "../py"
  branch:   "master"
  commit:   "12641cf"
git-subrepo:
  version:  "0.4.9"
  origin:   "???"
  commit:   "???"'.
ruff format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

1 file reformatted
```

Its at a non-recoverable state. If you want a clean history (of git subrepo clone), then you will need to (i) revert the change, (ii) manually disable your pre-commit hook, then (iii) perform the above clone, (iv) enable your pre-commit hook again, and then (v) perhaps peforms the code formatting in your subrepo to ensure consistency.

In my opinion, it is non-consistent, it leaves things in a dirty state, and intuitively, the hook that people expect to run (if at all) should be
- the ones inside the subrepo when you perform the clone/commit action
	```
	├── app
	│   ├── .pre-commit-config.yaml
	│   ├── subrepo1                            # <------ this was the git subrepo ops
	│   │  ├── .pre-commit-config.yaml
	```
- and not the ones at the main repo 
	```
	├── app                                      # <------ but this got executed
	│   ├── .pre-commit-config.yaml
	│   ├── subrepo1                            
	│   │  ├── .pre-commit-config.yaml
	```

See https://github.com/ingydotnet/git-subrepo/issues/193#issuecomment-259924850 for the prior discussion. Since the hooks inside subrepo cannot be the ones that actually get executed, I don't see why the default behaviour should be to verify

soraxas:
Hi thank you for the very detail write out (and including your reflection as well).

I don't agree partially with some of your view points. Points not explicitly mentioned below could be view as agreement with you:

> The git hooks are doing their job and should not allow you to accept code that doesn't pass their muster

This part is true

> As I said I have many repos that contain subrepos and both the subrepo and the containing repo have very aggressive hooks

The part about the containing repo has aggressive hooks has no bearing on the whole issue at all, as the inner hooks were never respected (I presume this part is clear to you, but mentioning them might muddle the focus on hooks that actually matters here (parent/root), especially to others stumble to this issue)

> [we think] ... hooks should be disabled because it is inconvenient when they fail on "3rd party code".

They are more than just inconvenient, they actually are confusing and for someone who don't fully understand the inner working of `gir-subrepo` (like me initially) they would have no idea what's wrong and dunno how to fix them. The contrived procedure that I mentioned is just me, somehow stumble my way, through me trial n error.

> ... It should not be store in a git config or .gitrepo file were it would effect users of the repo

Don't agree with it, with details below. But other methods like .gitattributes had always affect the user of the repo.
The current proposed method actually doesn't affect day-to-day committing of modfiled files in the subrepo. (i believe) it only affects the auto committing during initial clone, or pull from remote, which IMO, should never be verified as they represent the commit from remote. Modifying them with hook (without the initial commit) is already make the subrepo diverge from remote 

> It is possible to write hooks that behave the way you desire today without any infrastructure changes in git subrepo

It requires the changing of hooks, which could also be part of other well established infrastructure.

> Git does not have a mechanism to configure to turn off hooks for a specific repo, they rely on you doing that via the hooks.

This point is misleading, because hooks actually requires you to manually installed. Hence there's no "disabling" as enabling is not automatic. Not saying that I agree nor disagree whether hooks should be automatically enabled or not. But the argument that "official git has no way to declare disabling hooks, so git-subrepo should too" is invalid.

> Both the parent repository (the subrepo) and the containing repository hooks should be run.

I reckon this first depends on the philosophy of 

1. Viewing subrepo as just like a submodule where it should be treated as a separate module / 3rd part code (but without the submodule issue as elaborated in Q&A). And hence, all parent repo operation should skip the subrepo. And subrepo hook operation should operate in that subrepo. 
  i. Eg in normal git submodule, using lfs or other hooks has no bearing in the submodules.
ii. This is the viewpoint that brings the idea that hooks should be skipped

2. Subrepo is nothing special. It is just some subdirect where we happen to be able to push or pull from other remote.
  i. (perhaps this can be a valid viewpoint too, and I believe this is closer to your viewpoint)


-----

Thanks for the pointers on the exclude on solving pre-commit.config. It is indeed a solution, tho only for that kind of hooks

